PR #3, opened 2016-05-24
fixed some clippy warnings

llogiq:
This combines two if-conditions, pulls an else-if together and adds a bunch of Default impls.

### DIFF
--- a/src/parking_lot.rs
+++ b/src/parking_lot.rs
@@ -102,6 +102,12 @@ impl ThreadData {
     }
 }
 
+impl Default for ThreadData {
+    fn default() -> ThreadData {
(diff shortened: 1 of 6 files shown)


Amanieu:
You need a manual implementation of `Default` for the generic `ThreadParker`.

Amanieu:
I've fixed a few more clippy lints, and I think the rest just aren't worth fixing. In particular all those `Default` implementations are never going to be used since these types are internal to the library.

llogiq:
That's a good observation. Do you think we should reduce the scope of that lint to public types?

Amanieu:
Yes that would probably be a good idea.

PR #7, opened 2016-07-13
Fix doc on Mutex

jwilm:
`try_lock` returns an `Option`, not a `Result`.

---
This change is [Reviewable](https://reviewable.io/reviews/amanieu/parking_lot/7)


Amanieu:
Thanks!

PR #9, opened 2016-08-15
Make raw lock/unlock methods more versatile on Mutex and RwLock

whataloadofwhat:
Changes the raw lock and unlock methods to work on all `T: ?Sized`, not just `()`. The locking methods are somewhat useless I suppose but being able to unlock after using `mem::forget` on a Guard is useful.

Also removes `unsafe` on the `RwLock` raw locking methods. They are safe as far as I know (equivalent to `mem::forget(self.read())` and such), and there is no `unsafe` on the equivalent `Mutex` methods, so I don't think they are necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ in the Rust standard library:
     problem where all threads try to acquire the lock at the same time.
 12. `RwLock` supports atomically downgrading a write lock into a read lock.
 13. `Mutex` and `RwLock` allow raw unlocking without a RAII guard object.
-14. `Mutex<()>` and `RwLock<()>` allow raw locking without an RAII guard
+14. `Mutex<()>` and `RwLock<()>` allow raw locking without a RAII guard
     object.
 
 ## The parking lot
--- a/README.md
+++ b/README.md
@@ -54,8 +54,9 @@ in the Rust standard library:
     rest to wait on the associated `Mutex`. This avoids a thundering herd
     problem where all threads try to acquire the lock at the same time.
 12. `RwLock` supports atomically downgrading a write lock into a read lock.
-13. `Mutex<()>` and `RwLock<()>` allow raw locking and unlocking without a
-    RAII guard object.
+13. `Mutex` and `RwLock` allow raw unlocking without a RAII guard object.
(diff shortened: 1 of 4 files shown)


Amanieu:
I'm not too happy about making `raw_lock` and `raw_try_lock` available for all mutex types since they don't give you any way to access the underlying data. Maybe we could only expose `raw_unlock` to all mutex types while keeping `raw_lock` and `raw_try_lock` only for `Mutex<()>` (and possibly making them safe).

whataloadofwhat:
Keeping the raw locking to `<()>` is alright with me.

README.md
```diff
@@ -54,8 +54,9 @@ in the Rust standard library:
     rest to wait on the associated `Mutex`. This avoids a thundering herd
     problem where all threads try to acquire the lock at the same time.
 12. `RwLock` supports atomically downgrading a write lock into a read lock.
-13. `Mutex<()>` and `RwLock<()>` allow raw locking and unlocking without a
-    RAII guard object.
+13. `Mutex` and `RwLock` allow raw unlocking without a RAII guard object.
+14. `Mutex<()>` and `RwLock<()>` allow raw locking without an RAII guard
```
Amanieu: "a"

src/lib.rs
```diff
@@ -54,8 +54,9 @@
 //!     rest to wait on the associated `Mutex`. This avoids a thundering herd
 //!     problem where all threads try to acquire the lock at the same time.
 //! 12. `RwLock` supports atomically downgrading a write lock into a read lock.
-//! 13. `Mutex<()>` and `RwLock<()>` allow raw locking and unlocking without a
-//!     RAII guard object.
+//! 13. `Mutex` and `RwLock` allow raw unlocking without a RAII guard object.
+//! 14. `Mutex<()>` and `RwLock<()>` allow raw locking with a RAII guard
```
Amanieu: "without"

Ericson2314:
If the use-case is `mem::forget`, maybe a safer `put`/`take` interface like Haskell's `MVar` would be better? That requires an initialized bit so might be better as a separate type.

Amanieu:
Thanks!

PR #11, opened 2016-09-10
cast suseconds_t to c_long, fix build on x86_64-apple-darwin

Mrmaxmeier:
On `x86_64-apple-darwin` parking_lot_core currently fails to compile.
```
error[E0308]: mismatched types
   --> core/src/thread_parker/unix.rs:177:18
    |
177 |         tv_nsec: now.tv_usec * 1000,
    |                  ^^^^^^^^^^^^^^^^^^ expected i64, found i32
```
`libc::timespec.tv_nsec` is of type `libc::c_long` (i64 on x86_64). `libc::timeval.tv_usec` is of type `libc::suseconds_t` (i32 on x86_64). On i686 both types are i32 and the cast is a no-op.


Amanieu:
Thanks!

PR #22, opened 2017-02-11
Add a for_lock method on RwLockReadGuard and RwLockWriteGuard.

SimonSapin:
In Servo, I’m trying a scheme where mupltiple objects, each in `Arc<_>`, are protected by a single shared `RwLock`. This lock can be acquired once to access many objects.

https://bugzilla.mozilla.org/show_bug.cgi?id=1335941#c10

On each access, we assert that the guard being used indeed is for the same lock that protects the object being accessed. This is done by comparing `*const RwLock<()>` pointers. I currently keep this pointer in my guard wrapper types, but this is wasted space since this crate’s guards already hold a reference to the lock.

This commit adds public API to guards to expose this reference.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "parking_lot"
-version = "0.3.7"
+version = "0.3.8"
 authors = ["Amanieu d'Antras <amanieu@gmail.com>"]
 description = "More compact and efficient implementations of the standard synchronization primitives."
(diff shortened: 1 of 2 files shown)


SimonSapin:
If desired, I can add similar APIs to `MutexGuard` and `ReentrantMutexGuard` (though I have no plan to use those myself).

SimonSapin:
I’m going with a different approach (more convenient API) that requires storing that extra reference anyway in my guard wrapper types.

PR #33, opened 2017-05-07
Fix Compile Error on emscripten

CryZe:
This is not the root module, so you can't access std like that as a relative path. I'd also be pretty happy if we can have a 0.4.2 for this, as I'd like to publish the next version of my crate to crates.io, but it relies on this working ^^'


Amanieu:
Thanks!

Amanieu:
Version 0.4.2 is released with this fix!

CryZe:
Thank you :)

PR #34, opened 2017-05-30
Allow the usage of smallvec 0.4 and bump version

Eijebong:


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "parking_lot"
-version = "0.4.4"
+version = "0.4.5"
```
Amanieu: This isn't necessary, only the `parking_lot_core` version needs to be bumped.

core/Cargo.toml
```diff
@@ -9,7 +9,7 @@ repository = "https://github.com/Amanieu/parking_lot"
 keywords = ["mutex", "condvar", "rwlock", "once", "thread"]
 
 [dependencies]
-smallvec = "0.3"
+smallvec = ">= 0.3, < 0.5"
```
Amanieu: Just use `smallvec = "0.4"`.

Amanieu:
Thanks!

Eijebong:
Thanks for the quick reaction :)

PR #35, opened 2017-06-04
Use Duff's device and unroll spinwait loops

sstewartgallus:
This was benchmarked to give a small performance improvement from about 5.6kHz to 5.5kHz.

Old: parking_lot::Mutex   |      5.627 kHz |      5.684 kHz |      0.115 kHz
New: parking_lot::Mutex   |      5.525 kHz |      5.521 kHz |      0.225 kHz

It should also give a power usage improvement.

### DIFF
--- a/core/src/spinwait.rs
+++ b/core/src/spinwait.rs
@@ -42,40 +42,100 @@ fn thread_yield() {
 // using a hint to indicate to the CPU that we are spinning.
 #[cfg(all(feature = "nightly", any(target_arch = "x86", target_arch = "x86_64")))]
 #[inline]
-fn cpu_relax(iterations: u32) {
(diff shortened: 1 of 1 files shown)


sstewartgallus:
So, I think I was misreading the statistics. I did the tests again with 3 seconds.

Old: parking_lot::Mutex   |      5.461 kHz |      5.465 kHz |      0.111 kHz
new: parking_lot::Mutex   |      6.074 kHz |      6.098 kHz |      0.239 kHz

Now the results make more sense

jonhoo:
I guess this is related to #43 that I just filed. @sstewartgallus you probably know more about whether the new `std::sync::atomic::hint_core_should_pause` would be useful in this new PR than I do.

Issue #18, opened 2016-12-17
Deadlock detection


arthurprs:
How hard would it be to detect deadlocks and other types of "runtime bugs" in debug mode?

arthurprs:
This is interesting stuff. I guess that sort of instrumentation would also make possible to detect deadlocks due to poisoned locks as well.

Amanieu:
Poisoning is a different issue from deadlock detection: it's just a bit that you set in the `MutexGuard` destructor when panicking so that any future attempts to lock the mutex panic as well.

Amanieu:
I think this could be implemented as an additional feature on top of `ReentrantMutex`. We could have `lock` return an enum which indicates whether this is the first lock for the current thread. Deadlock detection could then be done by simply asserting that a recursive lock is never taken.

arthurprs:
That's one case, but I guess the most common one is like:

Thread 1 locks A, Thread 2 locks B
Thread 2 attempts to lock B and thread 2 attempts to lock A

Amanieu:
Unfortunately situations like that are much more difficult to detect. You would need a tool like [Hellgrind](http://valgrind.org/docs/manual/hg-manual.html).

arthurprs:
Just pointing it out. Maybe it's just better to leave it out to avoid confusion.

arthurprs:
Closed via #39

PR #39, opened 2017-08-28
Implement deadlock detection

arthurprs:
My stab to solve cc #18 using a wait graph. It requires more work but overall I think it's a good candidate for inclusion. Let me know your thoughts.

[x] feature flagged
[x] reasonably efficient
[nop] checking thread (automatic start?)
[x] reporting (how to?)
[ ] locking/unlocking in different threads break the algorithm

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: rust
 sudo: false
 
 rust:
-- 1.13.0
+- 1.18.0
 - stable
 - beta
 - nightly
@@ -18,7 +18,8 @@ script:
 - cd ..;
 - travis-cargo build
 - travis-cargo test
-- travis-cargo doc -- --no-deps -p parking_lot -p parking_lot_core
+- travis-cargo test -- --features=deadlock_detection
+- travis-cargo --only nightly doc -- --all-features --no-deps -p parking_lot -p parking_lot_core
 - if [ "$TRAVIS_RUST_VERSION" != "1.8.0" ]; then
       cd benchmark;
       travis-cargo build;
@@ -33,6 +34,7 @@ after_success:
 env:
   global:
   - TRAVIS_CARGO_NIGHTLY_FEATURE=nightly
+  - RUST_TEST_THREADS=1
 
 notifications:
   email: false
--- a/benchmark/Cargo.toml
+++ b/benchmark/Cargo.toml
@@ -18,3 +18,4 @@ path = "src/rwlock.rs"
 
 [features]
 nightly = ["parking_lot/nightly"]
+deadlock_detection = ["parking_lot/deadlock_detection"]
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -44,6 +44,13 @@
 extern crate smallvec;
 extern crate rand;
 
+#[cfg(feature = "deadlock_detection")]
+extern crate petgraph;
+#[cfg(feature = "deadlock_detection")]
+extern crate thread_id;
+#[cfg(feature = "deadlock_detection")]
+extern crate backtrace;
+
 #[cfg(unix)]
 extern crate libc;
 
@@ -77,3 +84,4 @@ pub use parking_lot::{ParkResult, UnparkResult, RequeueOp, UnparkToken, ParkToke
 pub use parking_lot::{DEFAULT_UNPARK_TOKEN, DEFAULT_PARK_TOKEN};
 pub use parking_lot::{park, unpark_one, unpark_all, unpark_requeue, unpark_filter};
 pub use spinwait::SpinWait;
+pub use parking_lot::deadlock;
--- a/core/src/parking_lot.rs
+++ b/core/src/parking_lot.rs
@@ -132,6 +132,14 @@ struct ThreadData {
 
     // ParkToken value set by the thread when it was parked
     park_token: Cell<ParkToken>,
+
+    // Is the thread parked with a timeout?
+    parked_with_timeout: Cell<bool>,
+
+    // Extra data for deadlock detection
+    // TODO: once supported in stable replace with #[cfg...] & remove dummy struct/impl
+    #[allow(dead_code)]
+    deadlock_data: deadlock::DeadlockData,
 }
 
 impl ThreadData {
@@ -149,6 +157,8 @@ impl ThreadData {
             next_in_queue: Cell::new(ptr::null()),
             unpark_token: Cell::new(DEFAULT_UNPARK_TOKEN),
             park_token: Cell::new(DEFAULT_PARK_TOKEN),
+            parked_with_timeout: Cell::new(false),
+            deadlock_data: deadlock::DeadlockData::new(),
         }
     }
 }
@@ -552,6 +562,7 @@ unsafe fn park_internal(key: usize,
     }
 
     // Append our thread data to the queue and unlock the bucket
+    thread_data.parked_with_timeout.set(timeout.is_some());
     thread_data.next_in_queue.set(ptr::null());
     thread_data.key.store(key, Ordering::Relaxed);
     thread_data.park_token.set(park_token);
@@ -574,6 +585,8 @@ unsafe fn park_internal(key: usize,
         Some(timeout) => thread_data.parker.park_until(timeout),
         None => {
             thread_data.parker.park();
+            // call deadlock detection on_unpark hook
+            deadlock::on_unpark(thread_data);
             true
         }
     };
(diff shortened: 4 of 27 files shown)


arthurprs:
I need to figure out how to test this properly.

arthurprs:
Ok, it's in good shape for review now.

core/src/parking_lot.rs
```diff
@@ -1035,3 +1042,67 @@ unsafe fn unpark_filter_internal(key: usize,
...
+            },
+        };
+    }
+}
+
+/// Returns true if a deadlock is detected
+#[cfg(feature = "deadlock_detection")]
+pub fn check_deadlock() -> bool {
+    // basic algorithm that builds a directed graph of the threads and the resources
+    // then checks for the presence of cycles (deadlocks)
```
Amanieu: While checking for deadlocks, you should lock the entire hash table, not just one bucket at a time. See `grow_hashtable` for how to do this properly. You might get false positives if you do not do this.

core/src/parking_lot.rs
```diff
@@ -1035,3 +1042,67 @@ unsafe fn unpark_filter_internal(key: usize,
...
+#[cfg(feature = "deadlock_detection")]
+pub fn deregister_resource(key: usize) {
+    unsafe {
+        let mut thread_data = None;
+        let thread_data = get_thread_data(&mut thread_data);
+        let resources = &mut (*thread_data.resources.get());
+        match resources.iter().rposition(|x| *x == key) {
+            Some(p) => resources.swap_remove(p),
+            None => {
+                check_deadlock();
```
Amanieu: Why do you call `check_deadlock` here? This code can only be reached if there is a bug in the implementation of `Mutex` (or whatever primitive calls `deregister_resource`).

core/src/parking_lot.rs
```diff
@@ -132,6 +135,9 @@ struct ThreadData {
 
     // ParkToken value set by the thread when it was parked
     park_token: Cell<ParkToken>,
+
+    // Resources
+    resources: UnsafeCell<Vec<usize>>,
```
Amanieu: The name "resources" isn't very informative. You should make it clearer that this is for deadlock detection.

core/src/lib.rs
```diff
@@ -77,3 +80,5 @@ pub use parking_lot::{ParkResult, UnparkResult, RequeueOp, UnparkToken, ParkToke
 pub use parking_lot::{DEFAULT_UNPARK_TOKEN, DEFAULT_PARK_TOKEN};
 pub use parking_lot::{park, unpark_one, unpark_all, unpark_requeue, unpark_filter};
 pub use spinwait::SpinWait;
+#[cfg(feature = "deadlock_detection")]
+pub use parking_lot::{register_resource, deregister_resource, check_deadlock};
```
Amanieu: I think the names `register_resource` and `deregister_resource` aren't very informative: it isn't immediately clear that this is related to deadlock detection. Maybe you could put these in their own submodule `parking_lot_core::deadlock_detector`.

Amanieu: It would also be nice if these functions we no-ops if the `deadlock_detection` feature is not defined. This avoids the need to spam `#[cfg(feature = "deadlock_detection")]` in the mutex implementations.

core/src/parking_lot.rs
arthurprs: It's a debug artifact, I'll remove it.

core/src/parking_lot.rs
arthurprs: I forgot to comment about this. I was thinking that the actual implementation could do a first-run like this and then do a second pass to confirm (locking all queues).

core/src/lib.rs
arthurprs: I have no idea how I came up with these, resources is a textbook parlance. I'll think of something.

arthurprs: Good idea.

Amanieu: I think that with the added submodule and some naming changes, it becomes clearer:

- `parking_lot_core::deadlock_detector::acquire_resource`
- `parking_lot_core::deadlock_detector::release_resource`

arthurprs:
I moved a lot of code around now, looks much better.

What's the most useful way of reporting a deadlock?

Amanieu:
> What's the most useful way of reporting a deadlock?

You should show a list of all the threads involved in the cycle, and print a backtrace for each thread. The backtrace can be done by waking each thread up one by one and using a special flag to indicate that it should print a backtrace and then go back to sleep.

arthurprs:
Do you think eprintln!() is enough for the first cut?

Amanieu:
I think that's probably fine for an initial implementation. Ideally you should return a `struct` which contains information about the cycle and a list of thread backtraces. It should be the job of the caller to actually decide how the information should be printed (or logged, etc).

src/deadlock.rs
```diff
@@ -0,0 +1,195 @@
...
+#[inline]
+pub fn acquire_resource(_key: usize) {
+    // noop
+}
+
+#[cfg(not(feature = "deadlock_detection"))]
+#[inline]
+pub fn release_resource(_key: usize) {
+    // noop
+}
```
Amanieu: I meant that you should make these functions no-ops in `parking_lot_core`, not here.

arthurprs: Ahh, makes sense.

core/src/parking_lot.rs
```diff
@@ -1040,112 +1091,253 @@ unsafe fn unpark_filter_internal(key: usize,
...
+    use std::sync::mpsc;
+    use std::sync::atomic::Ordering;
+    use std::collections::HashSet;
+    use backtrace::Backtrace;
+    use petgraph;
+    use petgraph::graphmap::DiGraphMap;
+
+    pub(super) unsafe fn on_unpark(td: &ThreadData) {
+        if td.deadlock_data.deadlocked.get() {
+            if let Some(sender) = (*td.deadlock_data.backtrace_sender.get()).take() {
```
Amanieu: This should never be `None`, so just use `unwrap()` here.

core/src/parking_lot.rs
```diff
@@ -1040,112 +1091,253 @@ unsafe fn unpark_filter_internal(key: usize,
...
+            }
+        }
+
+        for b in &(*table).entries[..] {
+            b.mutex.unlock();
+        }
+
+        // find cycles
+        let cycles = graph_cycles(&graph);
+        // println!("cycles {:?}", cycles);
```
Amanieu: Delete commented-out code.

core/src/parking_lot.rs
```diff
@@ -1040,112 +1091,253 @@ unsafe fn unpark_filter_internal(key: usize,
...
-            // double check with a full hash table lock
-            check_wait_graph_cycle(true)
-        } else {
-            false
+        unsafe {
+            // fast pass
+            if check_wait_graph_fast() {
+                // double check
+                let deadlocks = check_wait_graph_slow();
+                // TODO: call a configurable callback
```
Amanieu: Rather than using a callback, I would suggest returning a `struct` with all the backtraces of deadlocked threads.

core/src/parking_lot.rs
```diff
@@ -1040,112 +1091,253 @@ unsafe fn unpark_filter_internal(key: usize,
...
+        deadlock_impl::release_resource(_key);
+    }
+
+    /// TODO: doc
+    #[cfg(feature = "deadlock_detection")]
+    #[inline]
+    pub fn check_deadlock() -> bool {
+        deadlock_impl::check_deadlock()
+    }
+}
```
Amanieu: `acquire_resource` and `release_resource` should be unsafe functions. Consider what happens if we get a false-positive deadlock: the thread will be woken up and then will get stuck in `on_deadlock`. If it is unparked again, it will panic, which will leave the bucket in an inconsistent state.

core/src/lib.rs
```diff
@@ -46,6 +46,10 @@ extern crate rand;
 
 #[cfg(feature = "deadlock_detection")]
 extern crate petgraph;
+#[cfg(feature = "deadlock_detection")]
+extern crate thread_id;
```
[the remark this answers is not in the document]
arthurprs: The thing is that the std::thread IDs are not the system TIDs. I think in this case reporting the system TIDs is important to allow using further debug using gdb etc.

core/src/parking_lot.rs
arthurprs: Good point.

core/src/lib.rs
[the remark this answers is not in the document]
Amanieu: Ok, fair enough.

core/src/parking_lot.rs
arthurprs: Indeed, it's also easier.

src/deadlock.rs
```diff
@@ -1,4 +1,38 @@
...
+//!
+//! # Example
+//!
+//! ```
+//! #![feature(deadlock_detection)]
+//! use std::thread;
+//! use std::time::Duration;
+//! use parking_lot::deadlock;
+//!
+//! thread::spawn(move || {
```
Amanieu: Add a comment here: `// Create a background thread which checks for deadlocks every 10s`

Amanieu:
Ok, I think that's all. Just update that comment and I will merge it.

Thanks for all the great work! :+1:

arthurprs:
Done, should we do something about MutexGuard being Send'able?

Amanieu:
I really don't want to break the API for this, but it seems that deadlock detection is fundamentally impossible if we allow lock guards to be sent to other threads.

One alternative would be to make the deadlock detection resilient to false positives and simply add a disclaimer that deadlock detection may be unreliable if `MutexGuard`s are sent between threads. However this is tricky to implement since the current `ThreadParker` implementation is only designed to handle a single unparking thread.

arthurprs:
Let's add a disclaimer for now and in the future we can make the MutexGuard *not* Send'able when deadlock_detection is enabled.

arthurprs:
It wasn't as easy but I got it working.

src/deadlock.rs
```diff
@@ -33,10 +37,17 @@
...
+use std::marker::PhantomData;
+
 #[cfg(feature = "deadlock_detection")]
 pub use parking_lot_core::deadlock::check_deadlock;
 pub(crate) use parking_lot_core::deadlock::{acquire_resource, release_resource};
 
+#[cfg(not(feature = "deadlock_detection"))]
+pub(crate) struct DeadlockDetectionMarker(PhantomData<()>);
+#[cfg(feature = "deadlock_detection")]
+pub(crate) struct DeadlockDetectionMarker(PhantomData<*const ()>); // !Send
```
arthurprs: I made it a struct other than a type alias so that the user can get `DeadlockDetectionMarker` in the !Send error message.

Amanieu: Can you check that `MutexGuard<T>` is still `Sync` when `T: Sync`? Add a test for it similar to the `Send` test for the guards.

.travis.yml
```diff
@@ -32,7 +34,8 @@ after_success:
 
 env:
   global:
-  - TRAVIS_CARGO_NIGHTLY_FEATURE=nightly
+  - TRAVIS_CARGO_NIGHTLY_FEATURE=nightly,deadlock_detection
```
Amanieu: TRAVIS_CARGO_NIGHTLY_FEATURE should be left as it was before. The `deadlock_detection` feature should be explicitly tested separately.

.travis.yml
```diff
@@ -19,6 +19,8 @@ script:
 - travis-cargo build
 - travis-cargo test
 - travis-cargo doc -- --no-deps -p parking_lot -p parking_lot_core
+- travis-cargo --only nightly test -- --all-features
+- travis-cargo --only nightly doc -- --all-features --no-deps -p parking_lot -p parking_lot_core
```
Amanieu: Because of the way TRAVIS_CARGO_NIGHTLY_FEATURE is set, both tests are done with `deadlock_detection` entabled.

benchmark/Cargo.toml
```diff
@@ -18,3 +18,4 @@ path = "src/rwlock.rs"
 
 [features]
 nightly = ["parking_lot/nightly"]
+deadlock_detection = ["parking_lot/deadlock_detection"]
```
Amanieu: The benchmarks don't need deadlock detection.

.travis.yml
```diff
@@ -18,7 +18,9 @@ script:
 - cd ..;
 - travis-cargo build
 - travis-cargo test
+- travis-cargo test -- --features=deadlock_detection
 - travis-cargo doc -- --no-deps -p parking_lot -p parking_lot_core
+- travis-cargo --only nightly doc -- --all-features --no-deps -p parking_lot -p parking_lot_core
```
Amanieu: Why was this line added?

benchmark/Cargo.toml
arthurprs: It's not enabled by default, it's useful to measure the impact.

.travis.yml
arthurprs: Ahh, it should have replaced the one above

benchmark/Cargo.toml
Amanieu: Fair enough

Amanieu:
Do a `cargo fmt` to clean up the code style (for both parking_lot and parking_lot_core), and it should be good to go.

.travis.yml
arthurprs: It's to build the docs including all features, otherwise deadlock_detection will not show.

Amanieu: Then just add `--all-features` to the cargo doc line above.

arthurprs:
I squashed the deadlock detection changes, rustfmt changed quite a lot.

Amanieu:
Thanks for all the hard work! :+1:

PR #41, opened 2017-09-09
Make check_deadlock safe

arthurprs:
In the worst case the unparked deadlocked thread will panic (which is safe in rust definition). As far as I can see that this panic can't be triggered with safe code though. Even if lock guards are leaked (safe) there's no way of unparking the threads waiting on the leaked guard with safe code. As a compromise `acquire_/release_` are still marked unsafe though.

Plus minor doc improvements.


Amanieu:
Thanks!

PR #45, opened 2017-11-14
Add map methods to lock guards

Zoxc:
This makes it easier to port applications using RefCell to locks, since RefCell's Ref and RefMut have a map method.

### DIFF
--- a/src/mutex.rs
+++ b/src/mutex.rs
@@ -110,8 +110,9 @@ unsafe impl<T: Send> Sync for Mutex<T> {}
 /// `Deref` and `DerefMut` implementations.
 #[must_use]
 pub struct MutexGuard<'a, T: ?Sized + 'a> {
-    mutex: &'a Mutex<T>,
-    marker: PhantomData<(&'a mut T, DeadlockDetectionMarker)>,
+    raw: &'a RawMutex,
+    borrow: &'a mut T,
+    marker: PhantomData<DeadlockDetectionMarker>,
 }
 
 impl<T> Mutex<T> {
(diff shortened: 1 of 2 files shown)


src/rwlock.rs
```diff
@@ -165,6 +167,22 @@ impl<T> RwLock<T> {
 }
 
 impl<T: ?Sized> RwLock<T> {
+    fn read_guard(&self) -> RwLockReadGuard<T> {
```
Amanieu: Add `#[inline]`

src/rwlock.rs
```diff
@@ -165,6 +167,22 @@ impl<T> RwLock<T> {
...
+        RwLockReadGuard {
+            raw: &self.raw,
+            borrow: unsafe { &*self.data.get() },
+            marker: PhantomData,
+        }
+    }
+
+    fn write_guard(&self) -> RwLockWriteGuard<T> {
```
Amanieu: Add `#[inline]`

src/mutex.rs
```diff
@@ -143,6 +144,14 @@ impl<T> Mutex<T> {
 }
 
 impl<T: ?Sized> Mutex<T> {
+    fn guard(&self) -> MutexGuard<T> {
```
Amanieu: Add `#[inline]`

Amanieu:
Thanks! I'm just going to make a few small tweaks to the implementation and I will publish a new version of `parking_lot`.

Amanieu:
Published v0.5.0.

Issue #21, opened 2017-02-02
Upgrading a `RwLockReadGuard` to `RwLockWriteGuard` without dropping the lock.


ticki:
A function like `.upgrade()`.

Amanieu:
The problem with upgradable rwlocks is that they are very prone to deadlocks: consider the case where two reader threads both try to upgrade without dropping the lock.

While this could be supported by creating a third type of lock guard (`RwLockUpgradeGuard`) which allows concurrent readers but block writers and other upgradable locks, this significantly complicates the implementation.

ohanar:
Should this be closed now?

Amanieu:
Yes.

PR #48, opened 2017-11-24
Add upgradable guard to RwLock

ohanar:
This addresses #21.

### DIFF
--- a/src/raw_rwlock.rs
+++ b/src/raw_rwlock.rs
@@ -21,11 +21,13 @@ use deadlock;
 const TOKEN_SHARED: ParkToken = ParkToken(0);
 const TOKEN_EXCLUSIVE: ParkToken = ParkToken(1);
 
-const PARKED_BIT: usize = 1;
-const LOCKED_BIT: usize = 2;
-const SHARED_COUNT_MASK: usize = !3;
-const SHARED_COUNT_INC: usize = 4;
-const SHARED_COUNT_SHIFT: usize = 2;
+const PARKED_BIT: usize = 0b01;
+// A shared guard acquires a single guard resource
+const SHARED_GUARD: usize = 0b10;
+const GUARD_COUNT_MASK: usize = !PARKED_BIT;
+const GUARD_COUNT_SHIFT: usize = 1;
+// An exclusive lock acquires all of guard resource (i.e. it is exclusive)
+const EXCLUSIVE_GUARD: usize = GUARD_COUNT_MASK;
 
 pub struct RawRwLock {
     state: AtomicUsize,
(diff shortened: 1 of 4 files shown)


Amanieu:
Thanks, this looks great!

This is quite a lot of code, so I'll review it over the week-end. From a quick look so far, it looks mostly good though!

src/raw_rwlock.rs
```diff
@@ -13,19 +13,53 @@ use stable::{AtomicUsize, Ordering};
...
+        // specifies the number of usable bits, useful to test the
+        // implementation with fewer bits (the current implementation
+        // requires this to be at least 4)
+        const USABLE_BITS: usize = TOTAL_BITS;
+        (!0) >> (TOTAL_BITS - USABLE_BITS)
+    }
+    #[cfg(not(feature = "nightly"))] {
+        !0
+    }
+};
```
Amanieu: What is the purpose of this constant? It seems that it always has the value `!0`.

ohanar: I personally need to include this in a more complicated data structure (tagged pointers and the like...), so I wanted to make sure the implementation works and doesn't panic when using only a few bits for the state.

I added this so that it was easy for me to switch between testing with just a few bits and when using the full `usize`. In particular, everything should work perfectly fine with a value of `0b1111`, just with some different performance characteristics.

ohanar:
Yes, it definitely needs a look over as it is very tricky code. (and probably a few more tests wouldn't hurt)

src/raw_rwlock.rs
```diff
@@ -13,19 +13,53 @@ use stable::{AtomicUsize, Ordering};
...
+
+const PARKED_BIT: usize = 0b001;
+const UPGRADING_BIT: usize = 0b010;
+// A shared guard acquires a single guard resource
+const SHARED_GUARD: usize = 0b100;
+const GUARD_COUNT_MASK: usize = USABLE_BITS_MASK & !(SHARED_GUARD - 1);
+// An exclusive lock acquires all of guard resource (i.e. it is exclusive)
+const EXCLUSIVE_GUARD: usize = GUARD_COUNT_MASK;
+// An upgradable lock acquires just over half of the guard resource
+const UPGRADABLE_GUARD: usize = (GUARD_COUNT_MASK >> 1) + (SHARED_GUARD >> 1);
```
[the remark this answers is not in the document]
ohanar: It's currently correct. It should be `(GUARD_COUNT_MASK + SHARED_GUARD) >> 1` (i.e. just over half of the guards), but that addition might overflow. I'll add a comment to the source.

ohanar:
I've hopefully addressed all of your concerns, let me know if you find anything else concerning. I would be happy to squash the commits before merging.

Amanieu:
Well, I can't find anything wrong with the implementation, so I'll merge it. Thanks for the PR!

Amanieu:
Published in v0.5.2

PR #53, opened 2017-12-29
Fix future compatibilty warnings and bump version

MaloJaffre:
- Fix pulldown migration warnings in parking_lot_core
  
  Lists now require an empty line before them, and a link can no longer be
  split on multiple lines.
  See rust-lang/rust#44229.
  This is required to unblock rust-lang/rust#46278.

- Add an explicit type to silence a future compatibility warning
  
  `requeue_threads` now requires an explicit type because of the `is_null`
  method call later in the code.
  See rust-lang/rust#46906.

- Bump version to 0.2.9

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "parking_lot_core"
-version = "0.2.8"
+version = "0.2.9"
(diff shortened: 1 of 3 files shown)


Amanieu:
Thanks!

PR #57, opened 2018-01-31
Upgrade to rand 0.4

messense:
[rand changelog](https://github.com/rust-lang-nursery/rand/blob/master/CHANGELOG.md)


Amanieu:
Thanks!

PR #58, opened 2018-02-07
Add `?Sized` on `Send`- and `Sync`-impl for `Mutex` and `ReentrantMutex`

Lakelezz:
Would be awesome to add `?Sized` on `Mutex`'s `Send` and `Sync` in order to match it to https://doc.rust-lang.org/std/sync/struct.Mutex.html in that regard.


cuviper:
You might want to do the same for `ReentrantMutex`.

Amanieu:
Thanks for the PR!

PR #60, opened 2018-02-16
Bump minimum libc version requirement

staktrace:
In versions of libc prior to 0.2.27, libc::SYS_futex did
not exist. This is required by parking_lot_core when building
with the nightly feature enabled, like so:
    cargo +nightly build --features nightly

This patch bumps the minimum acceptable libc version so that
compiles under this configuration. This prevents parking_lot_core
from breaking builds when included in other projects where the
libc version is older than 0.2.27.


Amanieu:
Thanks!

PR #63, opened 2018-02-20
Provide correct implementation of AtomicElisionExt for x32

glaubitz:
The assembly implementation of AtomicElisionExt does not build on x32 as the x86_64 assembly code does not build there.

While it should be generally possible to write an assembly implementation of AtomicElisionExt for x32 later - and, FWIW, for all the other architectures - this change will at least fix the build failure of ```parking_lot``` on x32.

### DIFF
--- a/src/elision.rs
+++ b/src/elision.rs
@@ -90,7 +90,46 @@ impl AtomicElisionExt for AtomicUsize {
     }
 }
 
-#[cfg(all(feature = "nightly", target_arch = "x86_64"))]
+#[cfg(all(feature = "nightly", target_arch = "x86_64", target_pointer_width = "32"))]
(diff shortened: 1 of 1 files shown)


Amanieu:
Does changing `cmpxchgq` to `cmpxchgl` fix this by any chance?

glaubitz:
> Does changing cmpxchgq to cmpxchgl fix this by any chance?

Let me try ...

glaubitz:
Ok, I have updated the PR. ```cmpxchgl``` does work for x32 while ```cmpxchgq``` is the correct version for x86_64.

Amanieu:
This is now included in the latest version of parking_lot (0.5.4).

PR #67, opened 2018-04-23
Benchmarks now using barriers and support test iterations

Valloric:
There are two commits in this PR. 

The first one makes all benchmark threads wait on a common barrier before they start to reduce result skew from some threads starting earlier than others.

The second one adds support for running multiple benchmark iterations and collecting results across all runs.


Amanieu:
Thanks!

PR #70, opened 2018-04-27
Remove dependency on the `synchapi` winapi feature

alexcrichton:
Pulling in `parking_lot` to the Rust compiler unfortunately had adverse side
effects on the binary dependencies of rustc -- rust-lang/rust#49438. It turns
out this is due to a number of bugs happening all at once, and currently we're
looking for ways to fix the beta/master branches of rustc to fix this for now
(not necessarily in the best way).

This commit removes the activation of the `synchapi` feature of `winapi` which
prevents `libsynchronization.a` from being linked in (and prevents rustc from
accidentally picking up a dependency on a missing DLL). The `synchapi` feature
was only activated for access to the `Sleep` function, but that function
actually resides in `kernel32.dll`. As a result this commit inlines the
definition of `Sleep` into this crate to avoid needing to depend on an upstream
definition.

### DIFF
--- a/core/src/spinwait.rs
+++ b/core/src/spinwait.rs
@@ -17,11 +17,27 @@ use std::sync::atomic::spin_loop_hint;
 #[cfg(windows)]
 #[inline]
 fn thread_yield() {
+    // Note that this is manually defined here rather than using the definition
+    // through `winapi`. The `winapi` definition comes from the `synchapi`
+    // header which enables the "synchronization.lib" library. It turns out,
+    // however that `Sleep` comes from `kernel32.dll` so this activation isn't
+    // necessary.
+    //
+    // This was originally identified in rust-lang/rust where on MinGW the
+    // libsynchronization.a library pulls in a dependency on a newer DLL not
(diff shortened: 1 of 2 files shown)


core/src/spinwait.rs
```diff
@@ -17,11 +17,14 @@ use std::sync::atomic::spin_loop_hint;
 #[cfg(windows)]
 #[inline]
 fn thread_yield() {
+    extern "system" {
+        fn Sleep(a: winapi::shared::minwindef::DWORD);
+    }
```
Amanieu: Could you add a comment explaining why this is being pulled in directly?

alexcrichton:
Updated!

retep998:
Might be worth mentioning in the comment that this change was only done for the sake of building Rust itself as in any other situation we wouldn't be using the `libsynchronization.a` from MinGW.

alexcrichton:
Thanks @Amanieu! Could you also publish a new version so we can update rustc?

Amanieu:
Done!

alexcrichton:
Awesome, thanks @Amanieu!

PR #73, opened 2018-05-07
Fix Compile Error with x32 ABI

CryZe:
With the x32 ABI, the tv_nsec is specified to be a i64 explicitly. So casting to c_long causes a type conflict on the x86_64-unknown-linux-gnux32 target.


Amanieu:
I copied the code from linux.rs in 2309ce4. This should fix the issue.

PR #74, opened 2018-06-01
Remove unnecessary 'unsafe' blocks

mehcode:


Amanieu:
This will bump up the minimum required Rust version again, to 1.25. I would like to hold off on that for now since people have complained about it in the past.

mehcode:
Ah. Didn't realize. I just noticed the warning when I went to upgrade rand. Fair enough.

PR #77, opened 2018-06-07
Add "concurrency" category to Cargo.toml

lambda:
I was browsing [Crates.rs](https://crates.rs/index), an
alternative UI for crates.io, and expected to find parking_lot in
the "Concurrency" category, but I didn't since it looks like
parking_lot doesn't have any category set.  Add a category to aid
with discoverability.


Amanieu:
I was actually looking at crates.rs too and had the exact same thought!

Can you add a category to `parking_lot_core` as well?

lambda:
Sure, done.

Amanieu:
Thanks!

PR #79, opened 2018-06-18
Enable lock_api/nightly feature from parking_lot/nightly

myfreeweb:
Took me a while to figure out why the `const fn` functionality wasn't working...


Amanieu:
Nice catch!

PR #80, opened 2018-07-08
Remove depreacted rand::weak_rng(), in favor of SmallRng

bgianfo:
In https://github.com/rust-lang-nursery/rand/commit/1a4b3f753c41987e5077cadd5663758d56ffafd2
the weak_rng helper was deprecated, and SmallRng was
suggested to replace it instead.


Amanieu:
Ugh, this reminds me that I need to get around to fixing the CI scripts at some point.

PR #81, opened 2018-07-11
Escape [] in docs to avoid warnings on nightly

LukasKalbertodt:
Before this commit, these `[Experimental]` parts of the documentation
were seen as intra-links. Those links could not been resolved, resulting
in a warning on nightly.

It would be awesome, if this can be merged and maybe released as a point release, as `cargo doc` on nightly is quite noisy with many dependencies :)


Amanieu:
Thanks!

PR #98, opened 2018-10-13
Update to owning_ref 0.4

waywardmonkeys:


waywardmonkeys:
I realize this is a breaking change ... but wondering if it is worth it or not ...

Amanieu:
I don't think there is anyone actually using this feature...

It would require a major version bump for both `parking_lot` and `lock_api`.

PR #102, opened 2018-11-01
Add const_fn feature (more granular than nigthly)

niklasf:
[`min_const_fn`](https://github.com/rust-lang/rust/issues/53555) has been stabilized. It should still be opt-in, because it bumps the minimum Rust version to 1.31 (currently in beta).

This adds a new `const_fn` Cargo feature that allows using this without all current and future features implied by `nightly`. `nightly` still also implies `const_fn`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,8 +19,12 @@ rand = "0.5"
 [features]
 default = ["owning_ref"]
 owning_ref = ["lock_api/owning_ref"]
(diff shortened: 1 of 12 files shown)


niklasf:
Build failed due to:

```
error: trait bounds other than `Sized` on const fn parameters are unstable
   --> lock_api/src/rwlock.rs:236:6
    |
236 | impl<R: RawRwLock, T> RwLock<R, T> {
    |      ^
```

Looks like `min_const_fn` is not enough.

PR #105, opened 2018-11-20
Condvar: Return bool / usize from notify_one / notify_all

### LINKED ISSUE #104: Return bool from `Condvar::notify*`
From what I see, the methods could easily indicate whether somebody was waiting on the `Condvar`. Would you mind a PR for this?

torkleyy:
This simply returns the values already stored in `UnparkResult`.

Please tell me if you'd prefer not to merge the version bump.

## Changes

* Add return value to both methods to indicate if or how many threads were woken up (BREAKING)
* Bump version to 0.7.0 and add changelog entry
* Add `UnparkResult::requeued_threads`

## Issues

Fixes #104

### DIFF
--- a/src/condvar.rs
+++ b/src/condvar.rs
@@ -221,7 +221,7 @@ impl Condvar {
             };
             let res = parking_lot_core::unpark_requeue(from, to, validate, callback);
 
-            res.unparked_threads
+            res.unparked_threads + res.requeued_threads
         }
 
     }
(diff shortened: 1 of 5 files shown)


src/condvar.rs
```diff
@@ -198,8 +219,11 @@ impl Condvar {
                 }
                 TOKEN_NORMAL
             };
-            parking_lot_core::unpark_requeue(from, to, validate, callback);
+            let res = parking_lot_core::unpark_requeue(from, to, validate, callback);
+
+            res.unparked_threads
```
Amanieu: This will only ever be 1 or 0 since only 1 thread is unparked. The rest are requeued onto the mutex.

torkleyy: Sorry about that, should be fixed now. I added a new field to `UnparkResult` as well. Do you want me to seal the type to make it future proof?

torkleyy:
Please hold off with this one, I need to check something first.

torkleyy:
Okay, can be merged.

Amanieu:
Can you add a major version bump for parking_lot_core? This is a (minor) breaking change.

src/condvar.rs
Amanieu: What do you mean by "seal the type"?

torkleyy: Add a private field so the struct can't be exhaustively matched.

Amanieu: Ah yes, please do that.

Amanieu:
Thanks!

I want to make just one additional change to `notify_one` since we are bumping the version anyways, and then I will publish the new crates.

torkleyy:
Sounds great, thank you!

Amanieu:
I published the new version of parking_lot with your change.

torkleyy:
Thanks!

PR #106, opened 2018-12-01
Backport libstd style

faern:
Here are most of the changes I needed to make to the code in order to make it compile and run after copying it into libstd over in https://github.com/rust-lang/rust/pull/56410

### DIFF
--- a/src/condvar.rs
+++ b/src/condvar.rs
@@ -397,7 +397,7 @@ impl Condvar {
         guard: &mut MutexGuard<T>,
         timeout: Duration,
     ) -> WaitTimeoutResult {
-        // FIXME: Change to Intstant::now().checked_add(timeout) when stable.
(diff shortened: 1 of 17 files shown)


src/condvar.rs
```diff
@@ -387,12 +387,17 @@ impl Condvar {
...
+    /// # Panics
+    ///
+    /// Panics if the given `timeout` is so large that it can't be added to the current time.
     #[inline]
     pub fn wait_for<T: ?Sized>(
         &self,
         guard: &mut MutexGuard<T>,
         timeout: Duration,
     ) -> WaitTimeoutResult {
+        // FIXME: Change to Intstant::now().checked_add(timeout) when stable.
```
Amanieu: Should this `FIXME` also be added to `raw_mutex.rs` and `raw_rwlock.rs` where appropriate?

faern: Good idea. Fixed.

PR #107, opened 2018-12-01
Replace _padding with proper repr(align(x)) on newer Rust

faern:
This reduces the `Bucket` struct size from 112 bytes to 64 bytes when compiled on Rust >= 1.25 that supports `#[repr(align(x))]`

### DIFF
--- a/core/build.rs
+++ b/core/build.rs
@@ -2,6 +2,9 @@ extern crate rustc_version;
 use rustc_version::{version, Version};
 
 fn main() {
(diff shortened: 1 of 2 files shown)


faern:
To be clear. After applying this patch, the `Bucket` struct has size 112 and alignment 8 on Rust <1.25. And on newer Rust it has size 64 and alignment 64.

Amanieu:
Thanks!

PR #108, opened 2018-12-02
Safer word lock

faern:
I tried to reason a bit about the invariants of the unsafe code in WordLock. I think we can get away with them not being `unsafe` with some minor adjustments. Mostly this PR removes `unsafe` keywords from methods and adds `unsafe` blocks within them. But there is one more difference that I'm hoping should make this safe to call from the outside:
```rust
fn unlock(&self) {
    let state = self.state.fetch_and(!LOCKED_BIT, Ordering::Release);
    ...
}
```
Switching from `fetch_sub` to `fetch_and` is intended to make the method not corrupt `self.state` if called on an already unlocked `WordLock`. Previously multiple calls would continue to decrement the integer by one, locking and unlocking it as well as corrupting the part of it that is used as a pointer. By using the bit and operation this should just zero out the `LOCKED_BIT` without touching the rest. The test suite passes, I'm not sure if I have missed something else that makes this a bad change.

If you don't like the trait on `usize` providing the convenience methods we can skip them. They are unrelated to making the methods safe, other than it might be harder to screw code up if the bit manipulation is extracted.

### DIFF
--- a/core/src/word_lock.rs
+++ b/core/src/word_lock.rs
@@ -47,7 +47,7 @@ impl ThreadData {
 }
 
 // Returns a ThreadData structure for the current thread
-unsafe fn get_thread_data(local: &mut Option<ThreadData>) -> &ThreadData {
+fn get_thread_data(local: &mut Option<ThreadData>) -> &ThreadData {
     // Try to read from thread-local storage, but return None if the TLS has
     // already been destroyed.
     #[cfg(has_localkey_try_with)]
@@ -64,7 +64,7 @@ unsafe fn get_thread_data(local: &mut Option<ThreadData>) -> &ThreadData {
     if !cfg!(windows) && !cfg!(all(feature = "nightly", target_os = "linux")) {
         thread_local!(static THREAD_DATA: ThreadData = ThreadData::new());
         if let Some(tls) = try_get_tls(&THREAD_DATA) {
-            return &*tls;
+            return unsafe { &*tls };
         }
     }
 
--- a/core/src/word_lock.rs
+++ b/core/src/word_lock.rs
@@ -37,6 +37,7 @@ struct ThreadData {
 
 impl ThreadData {
     fn new() -> ThreadData {
+        assert!(mem::align_of::<ThreadData>() > !QUEUE_MASK);
         ThreadData {
             parker: ThreadParker::new(),
             queue_tail: Cell::new(ptr::null()),
@@ -46,8 +47,11 @@ impl ThreadData {
     }
 }
 
-// Returns a ThreadData structure for the current thread
-fn get_thread_data(local: &mut Option<ThreadData>) -> &ThreadData {
+// Invokes the given closure with a reference to the current thread `ThreadData`.
+fn with_thread_data<F, T>(f: F) -> T
+where
+    F: FnOnce(&ThreadData) -> T,
+{
     // Try to read from thread-local storage, but return None if the TLS has
     // already been destroyed.
     #[cfg(has_localkey_try_with)]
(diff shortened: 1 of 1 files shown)


Amanieu:
There is a good reason for using `fetch_sub` instead of `fetch_and`: it results in better code generation on x86 since it allows the use of the `xadd` instruction. `fetch_and` on the other hand requires a compare-exchange loop.

[Code](https://godbolt.org/z/0CTymD)

core/src/word_lock.rs
```diff
@@ -264,16 +266,45 @@ impl WordLock {
...
+
+trait LockState {
+    fn is_locked(&self) -> bool;
+    fn is_queue_locked(&self) -> bool;
+    fn queue_head(&self) -> *const ThreadData;
+    fn set_queue_head(&self, thread_data: *const ThreadData) -> Self;
+}
+
+impl LockState for usize {
+    #[inline(always)]
```
Amanieu: Plain `#[inline]` is probably sufficient here.

core/src/word_lock.rs
```diff
@@ -264,16 +266,45 @@ impl WordLock {
...
             break;
         }
     }
 }
+
+trait LockState {
+    fn is_locked(&self) -> bool;
+    fn is_queue_locked(&self) -> bool;
+    fn queue_head(&self) -> *const ThreadData;
+    fn set_queue_head(&self, thread_data: *const ThreadData) -> Self;
```
Amanieu: This doesn't modify `self`, so maybe `with_queue_head` is a better name?

faern:
Ah, crap. I suspected that was the reason. So I will make `lock` unsafe again and keep `lock_slow` safe. Still way less code under unsafe.

core/src/word_lock.rs
faern: That makes sense. Will also make every method consume `self` since it's just a `Copy` type anyway.

core/src/word_lock.rs
```diff
@@ -46,31 +47,40 @@ impl ThreadData {
...
+        panic::catch_unwind(AssertUnwindSafe(|| key.with(|x| f(x)))).ok()
     }
 
     // If ThreadData is expensive to construct, then we want to use a cached
     // version in thread-local storage if possible.
     if !cfg!(windows) && !cfg!(all(feature = "nightly", target_os = "linux")) {
         thread_local!(static THREAD_DATA: ThreadData = ThreadData::new());
-        if let Some(tls) = try_get_tls(&THREAD_DATA) {
-            return &*tls;
+        if let Some(result) = try_with_tls(&THREAD_DATA, &mut f) {
```
Amanieu: OK, so this is a bit subtle, but I would prefer if `f` was only called from one place and you simply extracted a reference from the thread-local storage. The reason for this is that if `f` is used in 2 different places, it will effectively be inlined twice into the function, which doubles the code size. This will require a bit of unsafe code to juggle the lifetimes.

faern: Ah, yeah. Makes sense. I did this now, but without much unsafe. Instead of juggling too many pointers I just stored the reference or owned value in a `Cow<ThreadData>`

core/src/word_lock.rs
```diff
@@ -48,39 +48,39 @@ impl ThreadData {
...
-        if let Some(result) = try_with_tls(&THREAD_DATA, &mut f) {
-            return result;
+        if let Some(tls_thread_data) = try_get_tls(&THREAD_DATA) {
+            thread_data_ptr = tls_thread_data;
         }
     }
+    if thread_data_ptr.is_null() {
+        // Otherwise just create a ThreadData on the stack
+        thread_data_storage = Some(ThreadData::new());
+        thread_data_ptr = thread_data_storage.as_ref().unwrap();
```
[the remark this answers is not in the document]
faern: I don't see how that would work here. `thread_data_storage` is always `None` when reaching this if block. So the closure to `get_or_insert_with` would always run and I would still need to get the address out for my pointer variable later. My initial declaration of `thread_data_storage` was a bit off, but I moved it now to better show how it's used.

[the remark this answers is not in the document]
Amanieu:
```
thread_data_ptr = thread_data_storage.get_or_insert_with(ThreadData::new);
```
Since `get_or_insert_with` returns a reference to the inner value in the option, you can avoid calling `as_ref` and `unwrap`.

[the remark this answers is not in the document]
faern: :man_facepalming: I completely missed it returns a reference. Of course it does.

Amanieu:
Thanks!

PR #109, opened 2018-12-02
Tighter defined hashtable and less unsafe functions

faern:
Some casting between `usize` and `*const HashTable` could be avoided by representing it as an `AtomicPtr<HashTable>` instead of `AtomicUsize`. Makes the types fit tighter to what they represent.

Also marked `get_hashtable` as safe, since I could not find any way it could have UB.

### DIFF
--- a/core/src/parking_lot.rs
+++ b/core/src/parking_lot.rs
@@ -194,7 +194,7 @@ impl Drop for ThreadData {
 }
 
 // Get a pointer to the latest hash table, creating one if it doesn't exist yet.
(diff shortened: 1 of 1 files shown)


Amanieu:
The historical reason for this is that `const fn` constructors were unstable, so it was not possible to have a static `AtomicPtr`. You had to emulate it with `AtomicUsize` and `ATOMIC_USIZE_INIT.

Since we require 1.24 now, it should be fine to just use `AtomicPtr`.

PR #110, opened 2018-12-06
Safer thread parkers

faern:
Continuing the work of trying to remove `unsafe` from methods/functions that actually can't result in UB. 

I know the spin based generic thread parker is also in the other PR (#106). But it was kind of needed here to get rid of unsafe `prepare_unpark` on (almost) all platforms. I will remove that commit from the other PR if/when this is merged.

I left the two constructors of the Windows thread parkers as unsafe. The documentation for `GetModuleHandleA` suggests it has to be carefully called and used in a multi threaded environment. At the same time I'm not sure what exactly could cause it to fail, or if that is even possible in our setup. If you have more information on the topic, maybe we can safely mark them as safe?


core/src/thread_parker/linux.rs
```diff
@@ -85,29 +72,40 @@ impl ThreadParker {
...
-                &ts,
+                ts_ptr,
+            )
+        };
+        debug_assert!(r == 0 || r == -1);
+        if r == -1 {
+            debug_assert!(
+                *libc::__errno_location() == libc::EINTR
+                    || *libc::__errno_location() == libc::EAGAIN
+                    || (!ts_ptr.is_null() && *libc::__errno_location() == libc::ETIMEDOUT)
```
Amanieu: `ts.is_some()` might be clearer here.

faern: Yes, I fully agree. fixed.

Amanieu:
The warnings about `GetModuleHandle` can be safely ignored I think. This only applies to DLLs that might be dynamically unloaded while we are still using them. This shouldn't apply to core system DLLs like ntdll.dll or api-ms-win-core-synch-l1-2-0.dll. However if you want to be absolutely sure then you can use `LoadLibraryA` instead, which is slower (but it should still be fine considering this is one-time initialization).

core/src/word_lock.rs
```diff
@@ -149,9 +149,12 @@ impl WordLock {
 
             // Get our thread data and prepare it for parking
             state = with_thread_data(|thread_data| {
+                #[cfg(all(unix, not(all(feature = "nightly", target_os = "linux"))))]
```
Amanieu: This feels a bit wierd, and is one of the reasons why I left the whole `ThreadParker` API as unsafe in the first place. At least add a comment explaining why this is necessary.

core/src/thread_parker/windows/mod.rs
```diff
@@ -40,13 +41,20 @@ impl Backend {
         }
 
         // Try to create a new object
```
Amanieu: So this isn't exactly your fault, but I just realized that this comment is nonsense. We've already created the `Backend`, now we just need to try to make it the global backend.

core/src/word_lock.rs
faern: Fixed

PR #111, opened 2018-12-15
No overflow panic on very long timeouts

faern:
Unstable support for checked_add on time types was finally merged (https://github.com/rust-lang/rust/pull/56490) so now we can avoid the possible panic that is `Instant::now() + timeout`.

First commit is unrelated. It's just some other simplification I thought it was nice to make.

### DIFF
--- a/src/condvar.rs
+++ b/src/condvar.rs
@@ -555,12 +555,20 @@ mod tests {
         let mut g = m.lock();
         let no_timeout = c.wait_for(&mut g, Duration::from_millis(1));
         assert!(no_timeout.timed_out());
+
         let _t = thread::spawn(move || {
             let _g = m2.lock();
             c2.notify_one();
         });
-        let timeout_res = c.wait_for(&mut g, Duration::from_millis(u32::max_value() as u64));
+        // Non-nightly panics on too large timeouts. Nightly treats it as indefinite wait.
+        #[cfg(feature = "nightly")]
+        let very_long_timeout = Duration::from_secs(u64::max_value());
+        #[cfg(not(feature = "nightly"))]
+        let very_long_timeout = Duration::from_millis(u32::max_value() as u64);
+
+        let timeout_res = c.wait_for(&mut g, very_long_timeout);
         assert!(!timeout_res.timed_out());
+
         drop(g);
     }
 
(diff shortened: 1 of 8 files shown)


src/condvar.rs
```diff
@@ -555,12 +558,20 @@ mod tests {
...
         let _t = thread::spawn(move || {
             let _g = m2.lock();
             c2.notify_one();
         });
-        let timeout_res = c.wait_for(&mut g, Duration::from_millis(u32::max_value() as u64));
+        // Non-nightly panics on too large timeouts. Nightly treats it as indefinite wait.
+        #[cfg(feature = "nightly")]
+        let very_long_timeout = Duration::from_secs(u64::max_value());
+        #[cfg(not(feature = "nightly"))]
+        let very_long_timeout = Duration::from_millis(u32::max_value() as u64);
```
Amanieu: This would look better with `if cfg!(feature = "nightly") {} else {}`.

src/util.rs
```diff
@@ -30,3 +32,13 @@ unsafe fn unreachable() -> ! {
         match *(1 as *const Void) {}
     }
 }
+
+#[inline]
+pub fn to_deadline(timeout: Duration) -> Option<Instant> {
+    #[cfg(feature = "nightly")]
+    let deadline = Instant::now().checked_add(timeout);
+    #[cfg(not(feature = "nightly"))]
+    let deadline = Some(Instant::now() + timeout);
```
Amanieu: Same here.

src/condvar.rs
faern: I instinctively always try to use `#[cfg]` to make the unwanted code go away as early as possible in the compile :)
I also think that won't work. Since `checked_add` does not exist on non-nightly. So it does not compile. `if false { checked_add() }` still requires `checked_add` even if it's never going to be executed.

faern: Sorry.. I misread which file the comment was on. Yes, here we can use `cfg!`. Fixed it.

faern:
The Linux CI error seems to be because `rustc_demangle` started using the rename dependency feature in the last release, coming from this commit: https://github.com/alexcrichton/rustc-demangle/commit/d74ae4b56ab6459f05afcb22c32371033cac75ff#diff-80398c5faae3c069e4e6aa2ed11b28c0

Some of the windows errors seems to be because `backtrace` started using the `#[repr(align(x))]` feature.

So maybe we need to bump the minimum version here anyway? Or lock these dependencies to very specific older patch releases. Requiring a newer compiler sounds very much like a breaking change to me, so I don't understand why these crates didn't bump the major version, but anyway.

PR #112, opened 2019-01-05
[WIP] New RwLock implementation

### LINKED ISSUE #101: Do upgradable reads block plain reads?
Hi! I see a behavior which I think does not match the docs, so either the docs are unclear, or there's a bug.

Here's the code that shows the behavior:

https://github.com/matklad/parking-lot-deadlock/blob/95989b3da40ed42e1ed7acba829d20b8b4319845/src/main.rs

If I run the code on my machine (linux, x86_64), I get a deadlock after a couple of hundred iterations.

I would expect that this program never deadlocks though, because, from my reading of the `upgradable_read` docs it follows that a `read` can never be blocked by a not-upgraded upgradable_read.

Amanieu:
The new implementation is inspired by boost's [`upgrade_mutex`](https://github.com/boostorg/thread/blob/fc08c1fe2840baeeee143440fba31ef9e9a813c8/include/boost/thread/v2/shared_mutex.hpp#L432) rwlock implementation.

The main benefit is that it fixes #101: upgradable reads no longer block normal reads. The code has also been refactored and is now much simpler than before.

cc @matklad @nikomatsakis

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ lock_api = { path = "lock_api", version = "0.1" }
 
 [dev-dependencies]
 rand = "0.6"
+lazy_static = "1.0"
 
 [features]
 default = ["owning_ref"]
--- a/lock_api/Cargo.toml
+++ b/lock_api/Cargo.toml
@@ -9,7 +9,7 @@ keywords = ["mutex", "rwlock", "lock", "no_std"]
 categories = ["concurrency", "no-std"]
 
 [dependencies]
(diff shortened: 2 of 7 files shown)


nikomatsakis:
@Amanieu we recently removed the use of `upgrade` from salsa, but I imagine @matklad might be able to test this code by reverting that PR, since they seemed pretty good at creating deadlocks before =)

Amanieu:
Appveyor failure is due to https://github.com/rust-random/rand/pull/722

PR #119, opened 2019-02-14
Make compatible as submodule to libstd

faern:
Preparing to include this repository as a git submodule under libstd over at https://github.com/rust-lang/rust/pull/56410

This is work in progress. Would like feedback, plus get confirmation from the CI that it works everywhere.

### DIFF
--- a/core/src/parking_lot.rs
+++ b/core/src/parking_lot.rs
@@ -5,6 +5,7 @@
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
+use super::libstd::time::{Duration, Instant};
 use super::thread_parker::ThreadParker;
 use super::util::UncheckedOptionExt;
 use super::word_lock::WordLock;
@@ -17,7 +18,6 @@ use core::{
 use rand::{rngs::SmallRng, FromEntropy, Rng};
 #[cfg(not(feature = "i-am-libstd"))]
 use smallvec::SmallVec;
-use std::time::{Duration, Instant};
 
 static NUM_THREADS: AtomicUsize = AtomicUsize::new(0);
 static HASHTABLE: AtomicPtr<HashTable> = AtomicPtr::new(ptr::null_mut());
--- a/core/src/parking_lot.rs
+++ b/core/src/parking_lot.rs
@@ -19,7 +19,26 @@ use rand::{rngs::SmallRng, FromEntropy, Rng};
 #[cfg(not(feature = "i-am-libstd"))]
 use smallvec::SmallVec;
 
+// When compiling the tests of std, there are basically two copies of std:
+// one normal one, and one with cfg(test). This means that during testing of
+// std there will be two instances of NUM_THREADS and HASHTABLE unless we make
+// sure both versions of std link to the same instance. It's usually not a
+// problem that different locks are based on different HASHTABLEs. Except if
+// the two locks both think they guard a global resource. Such as the mutexes
+// around the stdio file descriptors.
+// Here we use #[linkage] and #[export_name] to make both versions point to
+// the same instance.
+#[cfg_attr(all(test, feature = "i-am-libstd"), linkage = "available_externally")]
+#[cfg_attr(
+    feature = "i-am-libstd",
+    export_name = "_ZN3std16parking_lot_core11parking_lot11NUM_THREADSE"
+)]
 static NUM_THREADS: AtomicUsize = AtomicUsize::new(0);
+#[cfg_attr(all(test, feature = "i-am-libstd"), linkage = "available_externally")]
+#[cfg_attr(
+    feature = "i-am-libstd",
+    export_name = "_ZN3std16parking_lot_core11parking_lot9HASHTABLEE"
+)]
 static HASHTABLE: AtomicPtr<HashTable> = AtomicPtr::new(ptr::null_mut());
 
 // Even with 3x more buckets than threads, the memory overhead per thread is
--- a/core/src/parking_lot.rs
+++ b/core/src/parking_lot.rs
@@ -5,15 +5,17 @@
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
-use crate::thread_parker::ThreadParker;
-use crate::util::UncheckedOptionExt;
-use crate::word_lock::WordLock;
+use super::thread_parker::ThreadParker;
+use super::util::UncheckedOptionExt;
+use super::word_lock::WordLock;
 use core::{
     cell::{Cell, UnsafeCell},
     ptr,
     sync::atomic::{AtomicPtr, AtomicUsize, Ordering},
 };
+#[cfg(not(feature = "i-am-libstd"))]
 use rand::{rngs::SmallRng, FromEntropy, Rng};
+#[cfg(not(feature = "i-am-libstd"))]
 use smallvec::SmallVec;
 use std::time::{Duration, Instant};
 
@@ -85,6 +87,7 @@ struct FairTimeout {
     timeout: Instant,
 
     // Random number generator for calculating the next timeout
+    #[cfg(not(feature = "i-am-libstd"))]
     rng: SmallRng,
 }
 
@@ -93,6 +96,7 @@ impl FairTimeout {
     fn new(timeout: Instant) -> FairTimeout {
         FairTimeout {
             timeout,
+            #[cfg(not(feature = "i-am-libstd"))]
             rng: SmallRng::from_entropy(),
         }
     }
@@ -102,7 +106,11 @@ impl FairTimeout {
     fn should_timeout(&mut self) -> bool {
         let now = Instant::now();
         if now > self.timeout {
-            self.timeout = now + Duration::new(0, self.rng.gen_range(0, 1000000));
+            #[cfg(not(feature = "i-am-libstd"))]
+            let ns_offset = self.rng.gen_range(0, 1000000);
+            #[cfg(feature = "i-am-libstd")]
+            let ns_offset = 500000;
+            self.timeout = now + Duration::new(0, ns_offset);
             true
         } else {
             false
@@ -767,7 +775,10 @@ pub unsafe fn unpark_all(key: usize, unpark_token: UnparkToken) -> usize {
     let mut link = &bucket.queue_head;
     let mut current = bucket.queue_head.get();
     let mut previous = ptr::null();
+    #[cfg(not(feature = "i-am-libstd"))]
     let mut threads = SmallVec::<[_; 8]>::new();
+    #[cfg(feature = "i-am-libstd")]
+    let mut threads = Vec::new();
     while !current.is_null() {
         if (*current).key.load(Ordering::Relaxed) == key {
             // Remove the thread from the queue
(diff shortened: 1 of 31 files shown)


core/build.rs
```diff
@@ -2,10 +2,10 @@ extern crate rustc_version;
 use rustc_version::{version, Version};
 
 fn main() {
-    if version().unwrap() >= Version::parse("1.25.0").unwrap() {
+    if version().unwrap() >= Version::parse("1.25.0").unwrap() || cfg!(feature = "i-am-libstd") {
         println!("cargo:rustc-cfg=has_repr_align");
     }
-    if version().unwrap() >= Version::parse("1.26.0").unwrap() {
```
Amanieu: The build script won't be executed when the crate is built as a submodule.

lock_api/src/lib.rs
```diff
@@ -83,10 +83,11 @@
...
 //!   feature is `const fn` constructors for lock types.
 
-#![no_std]
+#![cfg_attr(not(feature = "i-am-libstd"), no_std)]
 #![warn(missing_docs)]
 #![cfg_attr(feature = "nightly", feature(const_fn))]
 
+#[cfg(not(feature = "i-am-libstd"))]
 #[macro_use]
 extern crate scopeguard;
```
Amanieu: I would prefer if you just included a minimal implementation of scopeguard directly in the crate as a module. Like [this](https://gist.github.com/Amanieu/a46426c7d5f2e6e76f8f8f9935830469). You would then be able to entirely drop the scopeguard dependency.

core/build.rs
faern: Ah, right. Was so much nicer to not need the `cfg(feature = "i-am-libstd")` everywhere in this code. But I guess I have to go back to that to get these features.

lock_api/src/lib.rs
faern: Ok. Will look at that.

faern:
@jethrogb `parking_lot` does not handle very well to crash in the parking code. Which is why all assertions in *all* `ThreadParker` implementations are just `debug_assert`s. So, we already don't unwind on errors, we just ignore them. What is the reason SGX can't tolerate this?

jethrogb:
SGX is an security isolation mechanism for running code on untrusted systems. Therefore, the program must detect when the system is trying to manipulate it into an invalid state. If any manipulation is detected, the program must abort and most definitely not try to continue normally. You can read more about these types of attacks in the [Iago attacks](https://hovav.net/ucsd/dist/iago.pdf) paper.

The error checks in the `thread_parker` implementation are different from the checks elsewhere in the code in that they detect a misbehaving system as opposed to a bug in the `parking_lot` code.

faern:
@jethrogb Could you evaluate my attempt to implement what you suggested over in the testing branch: https://github.com/faern/parking_lot/blob/as-libstd-submodule-test/core/src/thread_parker/sgx.rs#L23-L30 ?

I applied something very similar to your patch. But I used `std::process::abort` when not running as part of libstd instead of panicking. `std::process::abort` and the `rtabort!` macro end up calling the very same abort implementation, but `rtabort!` prints the message first.

Not sure it's a good idea to try to unwind or print anything when running outside libstd. Both unwinding and printing to stderr involves locking locks in the background. I suspect doing so if the parking impl is in an inconsistent state might either deadlock or double-panic.

jethrogb:
LGTM. But note that `panic` never locks stderr for printing. I agree that unwinding doesn't work well right now but IMO that should just be fixed in the implementation.

core/src/parking_lot.rs
```diff
@@ -81,36 +92,75 @@ impl Clone for Bucket {
...
+            FairTimeout { counter: 0 }
+        }
+
+        // Determine whether we should force a fair unlock, and update the timeout
+        #[inline]
+        pub fn should_timeout(&mut self) -> bool {
+            // On SGX it's expensive to call `Instant::now()`. So we resort
+            // to make every 100 call be true.
+            self.counter += 1;
+            self.counter % 100 == 0
```
faern: I just pulled this number out of thin air. Maybe we want the interval to be way different, or maybe it's sufficient to just always return `false`. But I had to start somewhere.

Amanieu: That feels very suspicious and should probably be reverted. Are you sure that `Instant::now` is a performance bottleneck. Keep in mind that we are only calling it on the slow path where we are performing a syscall to unpark a thread anyways.

faern: And every time we resize the hashtable. Which admittedly should not be very often after a few initial resizes. But we don't call it once per resize, but once per bucket per resize. However, those initial calls can be reduced to one per resize without actually changing this algorithm, that is true!

According to @jethrogb `Instant::now()` takes ~6us on SGX. As compared to ~100ns on Linux. On their machine. But I did not yet get any feedback as to whether or not this conditional compilation actually affected any benchmark. However, from debugging/tracing their runs of the test suite they saw the library being stuck within `Instant::now()` quite frequently. Mind, that was before the duplication bug for `HASHTABLE` was fixed, so maybe things are different now.

I have not yet had time to reinstall my lab machine so I can get SGX running and actually benchmark this myself.

core/src/parking_lot.rs
```diff
@@ -5,19 +5,31 @@
...
+#[cfg_attr(
+    feature = "i-am-libstd",
+    export_name = "_ZN3std16parking_lot_core11parking_lot11NUM_THREADSE"
+)]
 static NUM_THREADS: AtomicUsize = AtomicUsize::new(0);
+#[cfg_attr(all(test, feature = "i-am-libstd"), linkage = "available_externally")]
+#[cfg_attr(
+    feature = "i-am-libstd",
+    export_name = "_ZN3std16parking_lot_core11parking_lot9HASHTABLEE"
+)]
```
Amanieu: Can you add a comment explaining why this is needed?

core/src/thread_parker/sgx.rs
```diff
@@ -18,6 +17,18 @@ use std::{
     thread,
     time::Instant,
 };
+use core::sync::atomic::{AtomicBool, Ordering};
+
+
```
Amanieu: Run rustfmt to clean up stuff like this.

core/src/parking_lot.rs
faern: Done.

core/src/thread_parker/sgx.rs
faern: Done

src/remutex.rs
```diff
@@ -5,8 +5,9 @@
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
-use crate::raw_mutex::RawMutex;
-use lock_api::{self, GetThreadId};
+use super::libstd::mem;
```
Amanieu: `core::mem`

core/src/thread_parker/cloudabi.rs
```diff
@@ -5,13 +5,16 @@
...
 use cloudabi as abi;
+#[cfg(feature = "i-am-libstd")]
+use crate::sys::abi;
 use core::{
     cell::Cell,
     mem,
     sync::atomic::{AtomicU32, Ordering},
 };
-use std::{convert::TryFrom, thread, time::Instant};
+use super::libstd::{convert::TryFrom, thread, time::Instant};
```
Amanieu: `core::convert::TryFrom`

core/src/thread_parker/wasm.rs
```diff
@@ -9,7 +9,7 @@ use core::{
     arch::wasm32,
     sync::atomic::{AtomicI32, Ordering},
 };
-use std::{convert::TryFrom, thread, time::Instant};
+use super::libstd::{convert::TryFrom, thread, time::Instant};
```
Amanieu: `core::convert::TryFrom`

src/elision.rs
```diff
@@ -5,7 +5,7 @@
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
-use std::sync::atomic::AtomicUsize;
+use super::libstd::sync::atomic::AtomicUsize;
```
Amanieu: `core::sync::atomic::AtomicUsize`

core/src/parking_lot.rs
jethrogb: SGX usercalls are very slow and should be avoided if possible. If you turn every `send` call into two usercalls, you've now made waking up twice as slow (12µs instead of 6µs). However, this issue seems relatively minor compared to the next point.

SGX is a technology for running software in untrusted environments, by running that software in a so-called secure enclave. Any outside input (such as time, network data, etc.) must be assumed to be under the control of an adversary. By making scheduling decisions based on the return value of `insecure_time`, the attacker is getting fine-grained control over which tasks can make forward progress and which tasks can be held back. This is not acceptable. All scheduling decisions must be made by the code running in the enclave. The attacker may still interfere with scheduling, but it can't influence the ordering of synchronized events in the enclave.

What does this mean for `parking_lot`? I think `FairTimeout::should_timeout` should always return `false`. But more generally, it means I clearly don't understand the scheduling mechanisms in this crate enough to determine if it's secure. I'm not sure if I can approve swapping out the `std::sync` implementation for the SGX platform with this crate, without doing a more thorough review of the entire codebase.

faern: It's not doing it for every wake up. Only on the slow path. `parking_lot` has a fast and a slow path and a fast path. It always tries to use the fast path, but if that fails, it has to fall back to full OS parking etc.

The value returned from `should_timeout` does not dictate very much about scheduling in general at all. It just makes sure there is a fair lock handover approximately every 500µs. I think the worst thing the attacker could do is either disable the fairness (which your suggestion for a hardcoded `false` would also do) or slightly control when it's triggered. But it has no control at all over who will get the lock when it uses fairness vs not using fairness etc.

faern: > the attacker is getting fine-grained control over which tasks can make forward progress and which tasks can be held back

No they don't get this fine-grained control. As I described above, the worst thing they could do is control which of the unlock mechanisms are being used: fair or not fair. But they would not have any influence over what thread is next in line if they forced fairness, so they could not control who is going to take over the lock and continue executing. And in the case of a non fair unpark it's just a race between the unparked thread and any other thread that might be interested in the lock. The time does not affect the race in any way that I can think of.

dekellum:
Is there any newer version of this work, or any chance this one could be revived?  Or perhaps I just need some helpful advise. I imagine this PR includes making the parking_lot types (like `Mutex` and `Condvar`) API compatible with libstd 1.x.   Tokio might like to have a non-default feature that uses these types instead of the libstd ones.  If it was API compatible, that would be a lot easier!

Amanieu:
There are no plans to make parking_lot fully API-compatible with libstd. This PR does *not* do this.

The plan for libstd integration is for the libstd types to become wrappers around `RawMutex`, etc. But this all happens in libstd, not parking_lot.

dekellum:
Are there libstd PR's for that I could look it?  Thanks for your work any case!  I definitely have found it faster than libstd, in _blocking-permit_ crate I'm working on.

Amanieu:
You can have a look at https://github.com/rust-lang/rust/pull/56410 but I don't think it will help you much.

Amanieu:
Closing since this is not going to happen.

PR #122, opened 2019-03-04
Upgrade to rust2018

faern:
As we discussed on IRC. Let's upgrade all these three crates to Rust 2018 edition and get rid of some legacy conditional compilation etc.

Will greatly aid the work to integrate it into libstd over at #119

### DIFF
--- a/core/src/parking_lot.rs
+++ b/core/src/parking_lot.rs
@@ -5,6 +5,9 @@
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
+use crate::thread_parker::ThreadParker;
+use crate::util::UncheckedOptionExt;
+use crate::word_lock::WordLock;
 use rand::rngs::SmallRng;
 use rand::{FromEntropy, Rng};
 use smallvec::SmallVec;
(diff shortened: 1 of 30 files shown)


core/src/parking_lot.rs
```diff
@@ -175,14 +168,9 @@ where
 {
     // Try to read from thread-local storage, but return None if the TLS has
     // already been destroyed.
-    #[cfg(has_localkey_try_with)]
     fn try_get_tls(key: &'static LocalKey<ThreadData>) -> Option<*const ThreadData> {
         key.try_with(|x| x as *const ThreadData).ok()
     }
```
Amanieu: There isn't much point keeping that as a separate function now.

src/condvar.rs
```diff
@@ -668,7 +670,7 @@ mod tests {
         let mut g = m.lock();
         while !c.notify_one() {
             // Wait for the thread to get into wait()
-            ::MutexGuard::bump(&mut g);
+            crate::MutexGuard::bump(&mut g);
```
Amanieu: Just add `MutexGuard` to the import above.

src/once.rs
```diff
@@ -16,8 +16,8 @@ use std::sync::atomic::AtomicUsize as AtomicU8;
 type U8 = usize;
 use parking_lot_core::{self, SpinWait, DEFAULT_PARK_TOKEN, DEFAULT_UNPARK_TOKEN};
 use std::fmt;
```
Amanieu: `core::fmt`

core/src/parking_lot.rs
```diff
@@ -5,19 +5,20 @@
...
-use std::time::{Duration, Instant};
-use thread_parker::ThreadParker;
-use util::UncheckedOptionExt;
-use word_lock::WordLock;
+use core::{
+    cell::{Cell, UnsafeCell},
+    ptr,
+};
+use std::{
+    sync::atomic::{AtomicPtr, AtomicUsize, Ordering},
```
Amanieu: `sync::atomic` should be imported from `core`, not `std`.

core/src/parking_lot.rs
faern: True. Removed.

PR #124, opened 2019-03-22
Add wasm thread parker

faern:
This is in preparation for https://github.com/rust-lang/rust/pull/56410. To not make any platform regress from a good thread parking implementation to a spin lock based one, I add real `ThreadParker` implementations for them to `parking_lot`. Wasm's default locking primitives are really basic. I guess that is because wasm still does not have threads? But with the `target_feature = "atomic"` option it does utilize being able to wait on variables, so I thought we might want to do that here as well.

Ping @alexcrichton. You seem to be the one who implemented most of this atomic wasm thread parking logic in the standard library. So maybe you have some valuable input?

I'm submitting this PR before I have had the chance to even try it out myself. So you can stop me if I'm going in the completely wrong direction :)

### DIFF
--- a/core/src/thread_parker/wasm.rs
+++ b/core/src/thread_parker/wasm.rs
@@ -1,0 +1,118 @@
+// Copyright 2016 Amanieu d'Antras
+//
+// Licensed under the Apache License, Version 2.0, <LICENSE-APACHE or
+// http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. This file may not be
+// copied, modified, or distributed except according to those terms.
+
+use core::{
+    arch::wasm32,
+    mem,
+    sync::atomic::{AtomicUsize, Ordering},
+};
+use std::{convert::TryFrom, thread, time::Instant};
+
+// Helper type for putting a thread to sleep until some other thread wakes it up
+pub struct ThreadParker {
+    parked: AtomicUsize,
+}
+
+const UNPARKED: usize = 0;
+const PARKED: usize = 1;
+
+impl ThreadParker {
+    pub const IS_CHEAP_TO_CONSTRUCT: bool = true;
+
+    #[inline]
+    pub fn new() -> ThreadParker {
+        ThreadParker {
+            parked: AtomicUsize::new(UNPARKED),
+        }
+    }
+
+    // Prepares the parker. This should be called before adding it to the queue.
+    #[inline]
+    pub fn prepare_park(&self) {
+        self.parked.store(PARKED, Ordering::Relaxed);
+    }
+
+    // Checks if the park timed out. This should be called while holding the
+    // queue lock after park_until has returned false.
+    #[inline]
+    pub fn timed_out(&self) -> bool {
+        self.parked.load(Ordering::Relaxed) == PARKED
+    }
+
+    // Parks the thread until it is unparked. This should be called after it has
+    // been added to the queue, after unlocking the queue.
+    #[inline]
+    pub fn park(&self) {
+        while self.parked.load(Ordering::Acquire) == PARKED {
+            let val = unsafe {
+                wasm32::i32_atomic_wait(
+                    ptr(&self.parked),
+                    PARKED as i32, // we expect our thread is parked
+                    -1,            // wait infinitely
+                )
+            };
+            // we should have either woken up (0) or got a not-equal due to a
+            // race (1). We should never time out (2)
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -26,6 +27,7 @@ script:
 - travis-cargo --only stable test -- --features=deadlock_detection
 - travis-cargo --only beta test -- --features=deadlock_detection
 - travis-cargo --only nightly test -- --all --no-run --target x86_64-fortanix-unknown-sgx
+- travis-cargo --only nightly test -- --all --no-run --target wasm32-unknown-unknown
```
faern: I'm aware this does not even test with the atomic `target_future` at the moment... Will have to fix that before this would make sense to have even.

core/src/thread_parker/wasm.rs
```diff
@@ -0,0 +1,115 @@
+// Copyright 2016 Amanieu d'Antras
+//
+// Licensed under the Apache License, Version 2.0, <LICENSE-APACHE or
+// http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. This file may not be
+// copied, modified, or distributed except according to those terms.
+
+//! A simple spin lock based thread parker. Used on platforms without better
+//! parking facilities available.
```
bjorn3: I think this should be changed.

faern: Yes it should. Thank you. I just started out with the generic impl and iterated from there. Forgot some stuff :P

core/src/thread_parker/wasm.rs
```diff
@@ -0,0 +1,115 @@
...
+use core::{
+    arch::wasm32,
+    mem,
+    sync::atomic::{AtomicUsize, Ordering},
+};
+use std::{convert::TryFrom, thread, time::Instant};
+
+// Helper type for putting a thread to sleep until some other thread wakes it up
+pub struct ThreadParker {
+    parked: AtomicUsize,
```
faern: I'm not sure why this is not an `AtomicI32`. I just imitated what `libstd/sys/wasm/mutex_atomic.rs` does. The wait operation requires an `*mut i32`, and the `ptr()` method asserts that the sizes are equal. Maybe simply because this was written before `AtomicI32` was stabilized? If that's the case I guess this can be changed now.

alexcrichton:
Thanks for the ping @faern! I've got some general thoughts on the overall strategy with `parking_lot` in general, but here just for the wasm stuff this looks great to me. This won't actually be tested in upstream rust-lang/rust and requires a custom-built standard library to be tested at all. That being said it does look correct and appropriate for the wasm target, although that's just going by function names since I'm not very familiar with `parking_lot` and its primitives.

faern:
@alexcrichton Do you have any links/resources on how I can build such a custom standard library. So I can at least once verify locally that it compiles.

alexcrichton:
Sure yeah! You can follow the same steps as [one of our examples in wasm-bindgen](https://github.com/rustwasm/wasm-bindgen/blob/master/examples/raytrace-parallel/build.sh), which basically just uses xargo to build libstd with custom flags

faern:
I was able to get an atomics enabled libstd with xargo and compile the code. Which was a good thing, because there were a few minor errors to fix. Nothing that changes the algorithm or anything vital. Just `unsafe {}` around the `i32_atomic_*` calls and some casting of `PARKED` to `i32` in some places.

@alexcrichton Thanks for the help. Is there an actual wasm runtime with atomics support, or is this code not possible to run anywhere? :)

faern:
@Amanieu I added the `cfg_if!` macro, for selecting `thread_parker` implementation. Since the list of `cfg`s were getting long. Especially crazy was the last one that had to negate everything before it to include the generic parker. Do you think this is a good idea? This macro is available in libstd as well, so it won't cause a problem when integrated there it seems.

core/src/thread_parker/wasm.rs
```diff
@@ -0,0 +1,118 @@
...
+        // We don't need to lock anything, just clear the state
+        self.parked.store(UNPARKED, Ordering::Release);
+        UnparkHandle(&self.parked)
+    }
+}
+
+// Handle for a thread that is about to be unparked. We need to mark the thread
+// as unparked while holding the queue lock, but we delay the actual unparking
+// until after the queue lock is released.
+pub struct UnparkHandle<'a>(&'a AtomicUsize);
```
Amanieu: Use a raw pointer here. The target may no longer be valid if the thread has exited.

faern: Done

PR #125, opened 2019-03-26
Add Redox thread parker

faern:
This is in preparation for https://github.com/rust-lang/rust/pull/56410. To not make any platform regress from a good thread parking implementation to a spin lock based one, I add real `ThreadParker` implementations for them to `parking_lot`. It also simply makes `parking_lot` usable as a standalone library on Redox.

Ping @jackpot51. You seem to be the one who implemented most of the futex waiting logic in the standard library. So maybe you have some valuable input? Or you could ping someone who might be able to have a look at/help with this.

I started out with the existing `core/src/thread_parker/linux.rs` futex based implementation and basically just changed a few things to make the calls fit into how Redox expects them.

I have been able to verify that this builds, with cross compilation. But I have not yet been able to link it, nor run anything on a real Redox VM/machine. Will try to make that happen. But if someone with more Redox skills and an already set up system could help run the test suite, and maybe the benchmark sub-crate, that would be awesome.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ before_script:
   export PATH=$HOME/.local/bin:$PATH
 - if [ "${TRAVIS_RUST_VERSION}" = "nightly" ]; then
     rustup target add x86_64-fortanix-unknown-sgx;
+    rustup target add x86_64-unknown-redox;
   fi
 
 script:
@@ -26,6 +27,7 @@ script:
 - travis-cargo --only stable test -- --features=deadlock_detection
 - travis-cargo --only beta test -- --features=deadlock_detection
 - travis-cargo --only nightly test -- --all --no-run --target x86_64-fortanix-unknown-sgx
+- travis-cargo --only nightly build -- --all --target x86_64-unknown-redox
 - travis-cargo --only nightly doc -- --all-features --no-deps -p parking_lot -p parking_lot_core -p lock_api
 - cd benchmark
 - travis-cargo build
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -26,6 +27,7 @@ script:
 - travis-cargo --only stable test -- --features=deadlock_detection
 - travis-cargo --only beta test -- --features=deadlock_detection
 - travis-cargo --only nightly test -- --all --no-run --target x86_64-fortanix-unknown-sgx
+- travis-cargo --only nightly build -- --all --target x86_64-unknown-redox
```
Amanieu: Should this be `--no-run`?

faern: It can't even link. That's why I do `build` instead of `test --no-run`.

Amanieu:
The code looks good. I'd just like @jackpot51 to have a quick look before merging.

faern:
I have now been able to run the benchmarks on Redox. They work. Kind of. Sometimes it just never finishes, no results are printed. It just get stuck after printing the line
```
    name    |    average    |     median    | std.dev
```
But I get the exact same behavior with the generic spin lock thread parker as well. So I don't think the issue is with this parker.

I also did the same simple test as I [wrote about in the CloudABI PR](https://github.com/Amanieu/parking_lot/pull/126#issuecomment-478335551). A simple binary that locks a mutex twice. To see the CPU consumption of a thread that stays parked. I got the expected results. Without this PR: 100% CPU usage. With this PR: 0%.

I have not yet been able to run the full test suite.

jackpot51:
Sorry for not commenting earlier. The code looks good to me!

PR #126, opened 2019-03-30
Add CloudABI thread parker

faern:
Adding a proper `ThreadParker` implementation for the CloudABI platform, so it does not fall back on the spin lock based `ThreadParker`. This is to expand where `parking_lot` can be used, but also in preparation for https://github.com/rust-lang/rust/pull/56410.

The code in this PR is heavily inspired by, and basically just a mix of, [the existing pthread parker], the CloudABI [condvar] and [rwlock] implementations.

Ping @EdSchouten. Would you mind looking if this implementation seems sane? If you have the resources at hand, running the test suite would be very valuable as well.

I have only verified that this code compiles. I have not been able to run it yet. I will get some new hardware in a few days, then I might be in a better position to play with this for real.

[the existing pthread parker]: https://github.com/Amanieu/parking_lot/blob/56e313f8b23557da8b3b48fd30d3e9837c231ca7/core/src/thread_parker/unix.rs
[condvar]: https://github.com/rust-lang/rust/blob/26714219f1c5f8c748f95198d0824551ea2e0eeb/src/libstd/sys/cloudabi/condvar.rs
[rwlock]: https://github.com/rust-lang/rust/blob/26714219f1c5f8c748f95198d0824551ea2e0eeb/src/libstd/sys/cloudabi/rwlock.rs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ before_script:
 - if [ "${TRAVIS_RUST_VERSION}" = "nightly" ]; then
     rustup target add x86_64-fortanix-unknown-sgx;
     rustup target add x86_64-unknown-redox;
+    rustup target add x86_64-unknown-cloudabi;
   fi
 
 script:
@@ -28,6 +29,7 @@ script:
 - travis-cargo --only beta test -- --features=deadlock_detection
 - travis-cargo --only nightly test -- --all --no-run --target x86_64-fortanix-unknown-sgx
 - travis-cargo --only nightly build -- --all --target x86_64-unknown-redox
+- travis-cargo --only nightly build -- --all --target x86_64-unknown-cloudabi
 - travis-cargo --only nightly doc -- --all-features --no-deps -p parking_lot -p parking_lot_core -p lock_api
 - cd benchmark
 - travis-cargo build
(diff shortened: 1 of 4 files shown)


faern:
I pushed another commit. Where I try to make the code a bit nicer and hopefully harder to screw up. By introducing a `LockGuard`, working just like a `MutexGuard`. As such, unlocking will happen automatically and we can't accidentally wait on a condvar with an unlocked lock etc.

EdSchouten:
Hi Linus,

The patch looks very good. Thanks for working on this! I'll see if I can get this tested for you.

faern:
Thank you! As you might have seen in the code already, the new implementation requires nightly Rust, and also that you run it with `--features nightly`. Otherwise it just uses the generic parker.

If the test suite fails in any aspect, try without `--features nightly` just to check if it's the new thread parker that is to blame or not.

faern:
I now have a working CloudABI environment. I have been able to run the benchmarks. They work.

I also created a very simple binary that locks a mutex twice on a single thread. Thereby parking itself forever. Without `--features nightly` that program sits at 100% CPU usage on one core. And with this `ThreadParker` it stays at 0% CPU. So the parker seems to do what it should.

I have not been able to run the full test suite. Since a CloudABI Rust program needs to [set up stdout/err handles a bit special](https://cloudabi.org/write/rust/), and I have no idea how to do that for the normal Rust test runner.

EdSchouten:
> I now have a working CloudABI environment. I have been able to run the benchmarks. They work.

Ha! You beat me to it. Thanks!

PR #129, opened 2019-04-12
Fix was_last_thread value in the timeout callback of park()

Amanieu:
cc @faern

### DIFF
--- a/src/condvar.rs
+++ b/src/condvar.rs
@@ -675,4 +675,29 @@ mod tests {
         drop(g);
         t.join().unwrap();
     }
+
+    #[test]
+    fn test_issue_129() {
+        let locks = Arc::new((Mutex::new(()), Condvar::new()));
+
+        let (tx, rx) = channel();
+        for _ in 0..4 {
+            let locks = locks.clone();
+            let tx = tx.clone();
+            thread::spawn(move || {
+                let mut guard = locks.0.lock();
+                locks.1.wait(&mut guard);
+                locks.1.wait_for(&mut guard, Duration::from_millis(1));
+                locks.1.notify_one();
+                tx.send(()).unwrap();
+            });
+        }
+
+        thread::sleep(Duration::from_millis(100));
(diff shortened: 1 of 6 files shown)


faern:
We should probably add a test for the code that triggered this problem as well I feel? Since it was a real bug, but only one of 77000 crates made us find it :P

src/condvar.rs
```diff
@@ -675,4 +675,23 @@ mod tests {
...
+                let mut guard = locks.0.lock();
+                locks.1.wait(&mut guard);
+                locks.1.wait_for(&mut guard, Duration::from_millis(1));
+            });
+        }
+
+        thread::sleep(Duration::from_millis(100));
+        locks.1.notify_one();
+
+        thread::sleep(Duration::from_millis(1000));
```
faern: This test does not assert anything? I can't make this test fail on the master branch of `parking_lot`.

Simplest way to fix might be to add an mpsc channel into the mix and have the threads send a unit value over it after the two waits. Then make sure the correct number of channel items fall out in the main thread with `recv_timeout` instead of this final sleep.

faern: This seems to do the trick I think...

```rust
    #[test]
    fn test_issue_129() {
        let locks = Arc::new((Mutex::new(()), Condvar::new()));

        let (tx, rx) = channel();
        for _ in 0..4 {
            let locks = locks.clone();
            let tx = tx.clone();
            thread::spawn(move || {
                let mut guard = locks.0.lock();
                locks.1.wait(&mut guard);
                locks.1.wait_for(&mut guard, Duration::from_millis(1));
                locks.1.notify_one();
                tx.send(()).unwrap();
            });
        }

        thread::sleep(Duration::from_millis(100));
        locks.1.notify_one();

        for _ in 0..4 {
            assert_eq!(rx.recv_timeout(Duration::from_millis(10)), Ok(()));
        }
    }
```

Amanieu: Thanks!

PR #133, opened 2019-04-13
Add version detection for stabilized atomics and checked_add

faern:
Automatically detecting Rust >= 1.34 and make good use of the now stable `AtomicU{8,32}` types as well as `Instant::checked_add`.

This has been broken out from #119 since that PR is a bit large and this commit is not really related to the crate being used in libstd. We don't really know when the libstd stuff will be merged anyway, and this could be immediately useful to anyone using `parking_lot` on the latest stable Rust.

I'm just generally against "let's just put this in the already open PR, it'll be fine". I find it usually creates more problems than it solves.

### DIFF
--- a/src/once.rs
+++ b/src/once.rs
@@ -5,18 +5,21 @@
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
-#[cfg(feature = "nightly")]
-use std::sync::atomic::AtomicU8;
-use std::sync::atomic::{fence, Ordering};
-#[cfg(feature = "nightly")]
-type U8 = u8;
-#[cfg(not(feature = "nightly"))]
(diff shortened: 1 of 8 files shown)


src/once.rs
```diff
@@ -5,19 +5,22 @@
...
+#[cfg(not(has_sized_atomics))]
+use core::sync::atomic::AtomicUsize as AtomicU8;
+use core::{
+    fmt, mem,
+    sync::atomic::{fence, Ordering},
+};
 use parking_lot_core::{self, SpinWait, DEFAULT_PARK_TOKEN, DEFAULT_UNPARK_TOKEN};
 
+#[cfg(has_sized_atomics)]
+type U8 = u8;
```
faern: I split the type definitions from the imports. I find this much more readable.

PR #135, opened 2019-04-27
Serde

ckaran:
These are the modifications that I found necessary to make parking_lot work with serde.  It is feature gated behind the 'enable_serde' feature.  Please review it to make sure it is good, I'm still learning rust and may have done somethings poorly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,10 +14,6 @@ edition = "2018"
 parking_lot_core = { path = "core", version = "0.5" }
 lock_api = { path = "lock_api", version = "0.2" }
 
-# Optional dependency for supporting serde.  Use the feature "enable_serde" to
-# get access to this.
-serde = {version = "1.0.90", features = ["derive", "rc", "std"], optional = true}
-
 [dev-dependencies]
 rand = "0.6"
 lazy_static = "1.0"
@@ -33,7 +29,7 @@ default = []
 owning_ref = ["lock_api/owning_ref"]
 nightly = ["parking_lot_core/nightly", "lock_api/nightly"]
 deadlock_detection = ["parking_lot_core/deadlock_detection"]
-enable_serde = ["serde", "lock_api/enable_serde"]
+enable_serde = ["lock_api/enable_serde"]
 
 [workspace]
 exclude = ["benchmark"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ default = []
 owning_ref = ["lock_api/owning_ref"]
 nightly = ["parking_lot_core/nightly", "lock_api/nightly"]
 deadlock_detection = ["parking_lot_core/deadlock_detection"]
-enable_serde = ["lock_api/enable_serde"]
+serde = ["lock_api/serde"]
 
 [workspace]
 exclude = ["benchmark"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,9 +14,16 @@ edition = "2018"
 parking_lot_core = { path = "core", version = "0.5" }
 lock_api = { path = "lock_api", version = "0.2" }
 
+# Optional dependency for supporting serde.  Use the feature "enable_serde" to
+# get access to this.
+serde = {version = "1.0.90", features = ["derive", "rc", "std"], optional = true}
+
 [dev-dependencies]
 rand = "0.6"
 lazy_static = "1.0"
+
+# Used when testing out serde support.
+bincode = {version = "1.1.3"}
 
 [build-dependencies]
 rustc_version = "0.2"
@@ -26,6 +33,7 @@ default = []
 owning_ref = ["lock_api/owning_ref"]
 nightly = ["parking_lot_core/nightly", "lock_api/nightly"]
 deadlock_detection = ["parking_lot_core/deadlock_detection"]
+enable_serde = ["serde", "lock_api/enable_serde"]
 
 [workspace]
 exclude = ["benchmark"]
--- a/lock_api/Cargo.toml
+++ b/lock_api/Cargo.toml
@@ -13,5 +13,14 @@ edition = "2018"
 scopeguard = { version = "1.0", default-features = false }
 owning_ref = { version = "0.4", optional = true }
 
+# Optional dependency for supporting serde.  Use the feature "enable_serde" to
+# get access to this.
+serde = {version = "1.0.90", features = ["derive", "rc", "std"], optional = true}
+
+[dev-dependencies]
+# Used when testing out serde support.
+bincode = {version = "1.1.3"}
+
 [features]
 nightly = []
+enable_serde = ["serde"]
--- a/lock_api/Cargo.toml
+++ b/lock_api/Cargo.toml
@@ -15,7 +15,7 @@ owning_ref = { version = "0.4", optional = true }
 
 # Optional dependency for supporting serde.  Use the feature "enable_serde" to
 # get access to this.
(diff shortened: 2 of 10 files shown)


lock_api/src/rwlock.rs
```diff
@@ -230,6 +235,39 @@ pub struct RwLock<R: RawRwLock, T: ?Sized> {
...
+{
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        // FIXME It may be possible to use self.read() here instead, but using
+        // self.write() is the conservative choice.  If it *is* possible
+        // (meaning that race conditions are impossible, etc.), then we should
+        // use self.read() instead for performance reasons.
+        self.write().serialize(serializer)
```
Amanieu: Serde seems to use `read` here. I think think we should match the existing behavior of serde.

https://docs.rs/serde/1.0.90/src/serde/ser/impls.rs.html#544-557

lock_api/src/mutex.rs
```diff
@@ -93,6 +98,35 @@ pub struct Mutex<R: RawMutex, T: ?Sized> {
...
+impl<'de, R, T> Deserialize<'de> for Mutex<R, T>
+where
+    R: RawMutex,
+    T: Deserialize<'de> + ?Sized
+{
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        Deserialize::deserialize(deserializer).map(|value| Mutex::new(value))
```
Amanieu: You can just do `map(Mutex::new)`  here. Same for the other `Deserialize` implementations.

lock_api/Cargo.toml
```diff
@@ -13,5 +13,14 @@ edition = "2018"
 scopeguard = { version = "1.0", default-features = false }
 owning_ref = { version = "0.4", optional = true }
 
+# Optional dependency for supporting serde.  Use the feature "enable_serde" to
+# get access to this.
+serde = {version = "1.0.90", features = ["derive", "rc", "std"], optional = true}
```
Amanieu: Optional dependencies are already treated as feature names. So you can just use `cfg(feature = "serde)` and get rid of the "enable_serde" feature.

Cargo.toml
```diff
@@ -14,10 +14,17 @@ edition = "2018"
 parking_lot_core = { path = "core", version = "0.5" }
 lock_api = { path = "lock_api", version = "0.2" }
 
+# Optional dependency for supporting serde.  Use the feature "enable_serde" to
+# get access to this.
+serde = {version = "1.0.90", features = ["derive", "rc", "std"], optional = true}
```
Amanieu: `parking_lot` doesn't need to depend on `serde`. Only `lock_api` needs the dependency.

lock_api/Cargo.toml
Amanieu: Also, why are the "derive", "rc" and "std" features enabled? I don't think any of these are needed. In fact you should be using `default-features = false` so that no_std support for serde is enabled.

lock_api/src/rwlock.rs
ckaran: Works for me!  Like I said in the [original issue](https://github.com/Amanieu/parking_lot/issues/46#issuecomment-486461368), I was unsure as to what the best thing to do was, and was hoping you'd make the decision.

lock_api/Cargo.toml
ckaran: > Optional dependencies are already treated as feature names. So you can just use `cfg(feature = "serde)` and get rid of the "enable_serde" feature.

I didn't know that; I'm still learning rust, so thank you for teaching me something new!

> Also, why are the "derive", "rc" and "std" features enabled? I don't think any of these are needed. In fact you should be using `default-features = false` so that no_std support for serde is enabled.

I enabled them because they mirrored how I was using them in my own project.  As long as all those features can be enabled in a different project easily, then there's no need to enable them.  Would the normal method to enable them in a project that uses `parking_lot` be something like:
```toml
serde = {features=["parking_lot/serde/rc"]}
```
or would it be
```toml
serde = {features=["serde/rc"]}
```
?

As long as the method to enable the features is well-documented so that third party users can enable them as needed, then I think removing the requirements is a good idea.

Cargo.toml
ckaran: OK.  Do you want me to make all these changes, or do you want to do it and merge them back in?

ckaran:
Hi @Amanieu, I looked at your change suggestions, and they make sense to me.  Do you want me to make the changes, or would you prefer to make them yourself and merge them in?

Cargo.toml
ckaran: Yes, but if you don't know that, then you end up in the situation I was in with my project where all the tests from within the `lock_api` crate work, but once you pop up outside of it they don't because `parking_lot` isn't enabling `serde`.  By putting the optional feature into the top-level crate (`parking_lot`) new users are aware of the existence of the `enable_serde` feature, and can use it directly.

In addition, by putting `enable_serde` into `parking_lot`, you can turn it on and off on all of the dependent crates at the same time; this could be useful if you need to enable it for the `core` or `benchmark` crates at the same time.

lock_api/src/rwlock.rs
ckaran: Now changed.

lock_api/src/mutex.rs
ckaran: Now changed.

Cargo.toml
Amanieu: You can add a "proxy" serde feature to parking_lot, which just forwards to lock_api/serde:

```
[features]
serde = ["lock_api/serde"]
```

lock_api/Cargo.toml
Amanieu: If your own crate needs to use serde, then it will itself depend on serde and enable the features that it needs. Cargo will combine the serde features from all dependencies and enable all the ones that are needed.

In the case of lock_api, we only want to enable the bare minimum of features, so that we are maximally compatible. This is why we disable even the default serde features in lock_api so that we don't enable the "std" feature and can be compatible with no_std projects.

ckaran: > Cargo will combine the serde features from all dependencies and enable all the ones that are needed.

I didn't know that; thanks for letting me know!  Now changed.

Cargo.toml
ckaran: OK, I've added the proxy serde feature.  That said, I'm still using `enable_serde` and not `serde`.  Should I change that as well?  I think I remember seeing a comment online that if the name of a crate and a feature switch are the same, then there will be issues with Cargo, but I don't remember where I saw that... is that correct?

Amanieu: You should replace `enable_serde` with just `serde`. You don't need to add a `serde` feature to `lock_api` since optional crates will automatically create a feature with the same name as the crate.

ckaran: OK, I made the changes.  Thank you for explaining that about Cargo to me!

ckaran:
I just realized that there is one more thing to do; I haven't updated either the travis or appveyor test scripts to test the serde feature.  I'm unfamiliar with how to do with either, and will just be making educated guesses in updating those files; would you be willing to update them as needed to ensure the tests are done?

Amanieu:
Just add these 2 lines to .travis.yml:
```
- travis-cargo build -- --features serde
- travis-cargo test -- --features serde
```

Amanieu:
Thanks!

PR #137, opened 2019-05-03
Remove extern crate serde and don't wildcard import

faern:
Since we upgraded to Rust 2018 edition we don't need `extern crate`. I also felt it would be nicer to just import the four types we need into scope rather than a wildcard import.


faern:
Now you were too quick for me! I also realized I could remove an unused dependency.

PR #140, opened 2019-05-13
Fix feedback from libstd PR, part 1

faern:
Here I address some of the fairly low hanging fruit from https://github.com/faern/parking_lot/pull/1

1. Simplify `unlock_bucket_pair`. https://github.com/faern/parking_lot/pull/1/files#r283107487
2. Use `impl Trait` instead of a bunch of generics around the park/unpark calls. https://github.com/faern/parking_lot/pull/1/files#diff-f0c71d8c95c3701fb57bafd0a8198201R625
3. Just re-format a comment.
4. Remove all `#[inline(never)]`. https://github.com/faern/parking_lot/pull/1/files#diff-936abe3ecea632b68dddc509e87c7c39R142
4. Create a separate trait for the `ThreadParker` and `UnparkHandle` types. So we have a single place to stick the documentation. And to better ensure all implementers have the exact same API. https://github.com/faern/parking_lot/pull/1/files#r283155912

cc @alexcrichton

### DIFF
--- a/core/src/thread_parker/mod.rs
+++ b/core/src/thread_parker/mod.rs
@@ -52,7 +52,7 @@ pub trait UnparkHandleT {
 }
 
 cfg_if! {
-    if #[cfg(all(has_sized_atomics, target_os = "linux"))] {
+    if #[cfg(all(has_sized_atomics, any(target_os = "linux", target_os = "android")))] {
         #[path = "linux.rs"]
         mod imp;
     } else if #[cfg(unix)] {
--- a/core/src/thread_parker/mod.rs
+++ b/core/src/thread_parker/mod.rs
@@ -1,0 +1,86 @@
+use cfg_if::cfg_if;
+use std::time::Instant;
+
+/// Trait for the platform thread parker implementation.
+///
+/// All unsafe methods are unsafe because the Unix thread parker is based on
+/// pthread mutexes and condvars. Those primitives must not be moved and used
+/// from any other memory address than the one they were located at when they
+/// were initialized. As such, it's UB to call any unsafe method on
+/// `ThreadParkerT` if the implementing instance has moved since the last
+/// call to any of the unsafe methods.
+pub trait ThreadParkerT {
+    type UnparkHandle: UnparkHandleT;
+
+    const IS_CHEAP_TO_CONSTRUCT: bool;
+
+    fn new() -> Self;
+
+    /// Prepares the parker. This should be called before adding it to the queue.
+    unsafe fn prepare_park(&self);
+
+    /// Checks if the park timed out. This should be called while holding the
+    /// queue lock after park_until has returned false.
+    unsafe fn timed_out(&self) -> bool;
+
+    /// Parks the thread until it is unparked. This should be called after it has
+    /// been added to the queue, after unlocking the queue.
+    unsafe fn park(&self);
+
+    /// Parks the thread until it is unparked or the timeout is reached. This
+    /// should be called after it has been added to the queue, after unlocking
+    /// the queue. Returns true if we were unparked and false if we timed out.
+    unsafe fn park_until(&self, timeout: Instant) -> bool;
+
+    /// Locks the parker to prevent the target thread from exiting. This is
+    /// necessary to ensure that thread-local ThreadData objects remain valid.
+    /// This should be called while holding the queue lock.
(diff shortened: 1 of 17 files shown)


core/src/thread_parker/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+use std::time::Instant;
+
+/// Trait for the platform thread parker implementation.
+///
+/// All unsafe methods are unsafe because the Unix thread parker is based on
+/// pthread mutexes and condvars. Those primitives must not be moved and used
+/// from any other memory address than the one they were located at when they
+/// were initialized. As such, it's UB to call any unsafe method on
+/// `ThreadParkerT` if the implementing instance has moved since the last
+/// call to any of the unsafe methods.
```
faern: It's kind of sad how this one implementation spreads `unsafe` to all platforms and the API in general. If the generic Unix implementation just boxed the mutex/condvar I think it could become safe, right? It might lose some performance, but that can be measured so we have some numbers. If the difference is not too large, maybe it would be sane to make the entire thread parker API safe in this way?

We can probably move more platforms away from the generic Unix parker and to better, more specific parkers. For example, could not Android use the same futex based one as Linux? Is there nothing more modern than pthread mutexes for BSD?

Amanieu: I believe Android should already be using futexes? It uses the Linux kernel so `target_os` should be `linux`, right?

faern: No, it has `target_os = "android"`. See here: https://github.com/rust-lang/libc/blob/master/src/unix/mod.rs#L1141

core/src/thread_parker/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// released to avoid blocking the queue for too long.
+    ///
+    /// This method is unsafe for the same reason as the unsafe methods in
+    /// `ThreadParkerT`.
+    #[inline]
+    unsafe fn unpark(self);
+}
+
+cfg_if! {
+    if #[cfg(all(has_sized_atomics, target_os = "linux"))] {
```
Amanieu: You might as well add android here as well. Futex should work since Android also uses Linux.

core/src/thread_parker/mod.rs
bjorn3: Maybe use [`Pin`](https://doc.rust-lang.org/stable/std/pin/) to guarantee that self is not moved?

faern: @bjorn3 Do you have an example on how to do that? I don't know a way to express that without introducing a layer of indirection between the code holding the `ThreadParker` and the actual pthread C structs.

core/src/thread_parker/mod.rs
faern: Sure. I'm not able to test this myself, but now it's added.

core/src/thread_parker/mod.rs
alexcrichton: Note that there's a lot of UB scenarios with pthread utilities that are beyond movement (that's one of the major reasons for moving to parking_lot!). I suspect that this is is probably going to have to stay `unsafe` even if we were to box primitives, but it would be good to document all the unsafe guarantees we know of at least that must be upheld.

While it's unfortuante that the `unsafe` leaks to other platforms it's also sort of the reality of writing portable code. You're often limited by the lowest common denominator.

core/src/thread_parker/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+/// call to any of the unsafe methods.
+pub trait ThreadParkerT {
+    type UnparkHandle: UnparkHandleT;
+
+    const IS_CHEAP_TO_CONSTRUCT: bool;
+
+    fn new() -> Self;
+
+    /// Prepares the parker. This should be called before adding it to the queue.
+    unsafe fn prepare_park(&self);
```
alexcrichton: Could this and the methods below have expanded documentation as to why they're `unsafe`? (e.g. what has to be upheld to actually call them safely)

faern: I'll leave most of the requested documentation updates to @Amanieu. So that will likely be done in a separate PR.

faern:
@Amanieu Fixed the broken stuff. I think this is ready for review/merge now.

Amanieu:
Sorry that it took me so long to review this, but it looks mostly fine.

PR #141, opened 2019-05-20
Run latest rustfmt (1.2.2-nightly)

faern:
Apparently rustfmt changed the formatting. Running the latest version now yields this result... I can't say I like the new one better than the old. All of the`if`s that are used as expressions that became one liners are harder to read IMO. But if we want to be able to keep using a formatter at all we must adhere to it, or configure it with `rustfmt.toml` if we want to use it, but not get this result 🤷‍♂

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -39,16 +39,9 @@
 
 #![warn(missing_docs)]
 #![warn(rust_2018_idioms)]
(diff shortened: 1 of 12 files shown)


faern:
The CI is broken. This PR only changes formatting.

Amanieu:
I'm just going to merge it manually.

Issue #138, opened 2019-05-07
Shall `rand` be a dev-dependency in parking_lot_core, like in parking_lot?


vi:
Just connecting `parking_lot` brings entire chunk of `rand_*` entries into `cargo tree`.

Amanieu:
This is by design, some randomness is used as part of the fairness mechanism in parking_lot_core.

matklad:
Note that the future version of rand plans to prune dependencies a bit: https://github.com/rust-random/rand/issues/715

faern:
Could we not swap the `rand` dependency for `rand_pcg`? `parking_lot` currently uses `SmallRng` which in the background is just `rand_pcg::Pcg64Mcg` or `rand_pcg::Pcg32` depending on architecture.

matklad:
Another option, if we don't need super-high-quality randomness, is to just vendor xor-shift RNG, like the `[T]::sort` does:

https://github.com/rust-lang/rust/blob/c28084ac16af4ab594b6860958df140e7c876a13/src/libcore/slice/sort.rs#L486-L493

faern:
If we include a simple PRNG in `parking_lot` itself we get the benefit of being able to use it when `parking_lot` acts as the locking implementation inside libstd (#119) and we get less conditional compilation and a bit more randomness to the fairness in libstd.

Amanieu:
Vendoring a simple PRNG sounds like a good idea that will solve multiple problems.

faern:
I copied the linked PRNG from libstd and took a stab at implementing this. See #144. I have not evaluated the quality of that PRNG or anything. But this compiles and the tests pass.

vi:
> Vendoring a simple PRNG

What about vendoring a simple PRNG into Rust's libcore? It doesn't look OK if everybody around starts vendoring a simple PRNG.

Amanieu:
No, crates that need a PRNG should just use the `rand` crate. We only need to vendor one here because `parking_lot` is intended to be used as part of libstd, which cannoy depend on other crates.

matklad:
It might be worthwhile to do a release of `parking_lot` without `rand`: `rand` recieved a major version bump recently, so published `parking_lot` now depends on outdated version.

Amanieu:
I just published 0.8.1 without rand.

PR #144, opened 2019-05-30
Implement own xorshift PRNG to avoid rand dependency

faern:
Trying to solve #138 

The PRNG implementation is stolen from https://github.com/rust-lang/rust/blob/c28084ac16af4ab594b6860958df140e7c876a13/src/libcore/slice/sort.rs#L486-L493 as pointed out in the issue.

If we are happy with the PRNG algorithm, the only question is the seed. I simply used a pointer readily available when creating the `FairTimeout`. We could potentially also use the nanos offset from the elapsed timeout as a seed. Something like:
```rust
let seed = now.duration_since(self.timeout).as_nanos() as u32;
let nanos = Self::gen_u32_with_seed(seed) % 1_000_000;
```

### DIFF
--- a/core/src/parking_lot.rs
+++ b/core/src/parking_lot.rs
@@ -13,7 +13,6 @@ use core::{
     ptr,
     sync::atomic::{AtomicPtr, AtomicUsize, Ordering},
 };
-use rand::{rngs::SmallRng, FromEntropy, Rng};
 use smallvec::SmallVec;
 use std::time::{Duration, Instant};
 
@@ -80,14 +79,14 @@ struct FairTimeout {
     // Next time at which point be_fair should be set
     timeout: Instant,
 
-    // Random number generator for calculating the next timeout
-    rng: SmallRng,
+    // the PRNG state for calculating the next timeout
+    seed: u32,
 }
 
 impl FairTimeout {
     #[inline]
     fn new(timeout: Instant) -> FairTimeout {
-        FairTimeout { timeout, rng: SmallRng::from_entropy() }
+        FairTimeout { timeout, seed: &timeout as *const _ as usize as u32 }
     }
 
     // Determine whether we should force a fair unlock, and update the timeout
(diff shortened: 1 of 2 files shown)


faern:
A seed of zero will generate the output zero. If that ever happened we would be stuck forever on zero. Luckily the only input that gives zero as output is zero. So we must make sure the initial seed is non-zero and we are fine.

core/src/parking_lot.rs
```diff
@@ -80,27 +79,37 @@ struct FairTimeout {
...
-    rng: SmallRng,
+    // the PRNG state for calculating the next timeout
+    seed: u32,
 }
 
 impl FairTimeout {
     #[inline]
     fn new(timeout: Instant) -> FairTimeout {
-        FairTimeout { timeout, rng: SmallRng::from_entropy() }
+        FairTimeout { timeout, seed: &timeout as *const _ as usize as u32 }
```
Amanieu: I'm not a big fan of using memory addresses as seeds since they could in theory be used to leak memory layout information to attackers and defeat address space randomization. Could we maybe use `timeout.nanos() + 1` as the initial seed instead?

faern: Sure. I just now realized other downsides with the address approach as well. Fixing :+1:

faern: Well, `Instant` has no `nanos` method. We would need to compute a `Duration` for that.

faern: We could just take the `seed` as argument and have `HashTable::new` just have an incrementing counter it sends. So first `FairTimeout` has seed `1` etc.

Amanieu: Yes, an incrementing counter would work fine.

faern:
I removed the word "random" from "Random time between 0 and 1ms.". This is now very deterministic so I did not want to use the word in the wrong way.

faern:
I will be offline from now until Monday.

PR #148, opened 2019-06-03
Fix feedback from libstd, part 2

faern:
Continuation of #140.

Picking misc issues from the review over at https://github.com/faern/parking_lot/pull/1 and fixing them.

### DIFF
--- a/lock_api/src/mutex.rs
+++ b/lock_api/src/mutex.rs
@@ -172,6 +172,7 @@ impl<R: RawMutex, T: ?Sized> Mutex<R, T> {
     #[inline]
     pub fn lock(&self) -> MutexGuard<'_, R, T> {
         self.raw.lock();
+        // SAFETY: The lock is held, as required.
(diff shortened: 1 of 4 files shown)


lock_api/src/mutex.rs
```diff
@@ -169,7 +172,7 @@ impl<R: RawMutex, T: ?Sized> Mutex<R, T> {
     #[inline]
     pub fn lock(&self) -> MutexGuard<'_, R, T> {
         self.raw.lock();
-        self.guard()
+        unsafe { self.guard() }
```
Centril: It would be good to leave a comment on each new introduced `unsafe {` discharging all proof obligations. In this case something like:
```
// SAFETY: The lock is held, as required.
```
While it is in the same file, things could be shuffled around with refactorings and it would be good to not lose these reasonings.

faern: Done

PR #151, opened 2019-06-09
Use hint::unreachable_unchecked instead of Void enum

taiki-e:
[`hint::unreachable_unchecked`](https://doc.rust-lang.org/nightly/core/hint/fn.unreachable_unchecked.html) stabilized in 1.27.


Amanieu:
Thanks!

PR #159, opened 2019-07-11
Staging for parking_lot 0.9.0,  _core 0.6.0 and lock_api 0.3.1 releases

dekellum:
This includes all necessary version changes and overhauls the CHANGELOG to show more recent releases for all three crates (_parking_lot_, _parking_lot_core_, _lock_api_). Before releasing, the top  `(TBD)` should be replaced with the current date and tags for all crates added.    I also request that you add the following tags to this repo retroactively:

``` bash
git tag core-0.5.0 3259fd72e0d76e9c9a6f31ff41a996648fe9ccfa
git tag lock_api-0.1.5 2653619a613bc12dd6ab486f789d98956b65cdf0
git tag lock_api-0.2.0 3259fd72e0d76e9c9a6f31ff41a996648fe9ccfa
git tag lock_api-0.3.0 ace8280969b6362c3f3753506d90122c358de580
```

Fixes #158

PR #160, opened 2019-07-11
Upgrade rand dev dependency to 0.7

dekellum:
The latest rand.  Also avoids some annoying crate duplicates for dev. Should not effect end users.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ env:
 
 matrix:
   include:
-    - rust: 1.31.0
+    - rust: 1.32.0
       os: linux
       script: &script
         - cargo build
@@ -49,7 +49,7 @@ matrix:
         - cargo run --release --bin rwlock -- 1 1 1 0 1 2
         - cd ..
 
-    - rust: 1.31.0
+    - rust: 1.32.0
       os: osx
       script: *script
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ parking_lot_core = { path = "core", version = "0.5" }
 lock_api = { path = "lock_api", version = "0.3" }
 
 [dev-dependencies]
-rand = "0.6"
+rand = "0.7"
 lazy_static = "1.0"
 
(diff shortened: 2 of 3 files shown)


Amanieu:
I think that you need to bump the minimum Rust version to 1.32 to fix the CI failures.

dekellum:
Thanks, I didn't look at CI given the prior failures, but rand 0.7's MSRV is a bigger issue. We can (SemVer wise) bump MSRV and include with #159, though now I'm not so sure rand 0.7 is worth a MSRV bump for this release. 

BTW: Looking at CI now, below linked failure isn't explained by MSRV?

https://travis-ci.org/Amanieu/parking_lot/jobs/557506159#L299 

Is this the issue fixed by your recent da16c2c7?

Amanieu:
Bumping the MSRV is not an issue since we are going to do breaking changes to parking_lot and parking_lot_core anyways.

And yes, that CI failure should be fixed by da16c2c.

dekellum:
My only reservation is that its kind of a soft need (rand 0.7 and MSRV 1.32 via dev dependency for testing only). I've gone ahead and updated it here though.

Amanieu:
Alternatively, I guess we could keep 1.31 as the MSRV and simply not run tests on 1.31, only build checks. We should probably have tests on 1.32 in that case.

dekellum:
I could clarify this nuance in the README section for MSRV and CHANGELOG, e.g.: tests now require 1.32, will likely continue working for users on 1.31. Would do that in #159 if you choose to merge this PR. Alternatively you can just keep this PR around until you have a more compelling need for 1.32. :)

dekellum:
As of 79a007d, CI is back to just the unrelated *-windows-gnu target failures only.

Amanieu:
I just fixed the Windows CI, but your changes to appveyor.yml are conflicting.

dekellum:
Rebased, keeping your version of appveyor.yml (with 1.32) verbatim.  CI now passes, thanks.

PR #162, opened 2019-07-15
MSRV back to 1.31

matklad:


matklad:
not sure why travis build won't start...

.travis.yml
```diff
@@ -7,6 +7,16 @@ env:
 
 matrix:
   include:
+    # can't run tests on 1.31.0 due to rand dev-dependency
+    - rust: 1.31.0
+      os: linux
+      script
```
Amanieu: You are missing a `:` here. This is probably why the Travis build isn't working.

matklad:
should be good now!

PR #164, opened 2019-07-20
Fix compile errors on Windows due to old cfg_if version

rkarp:
`cfg-if 0.1.4` and below don't work correctly on Windows. I get a compile error on `parking_lot_core` due to another crate in the build forcing version `0.1.3`.


Amanieu:
Published in parking_lot_core 0.6.2.

PR #167, opened 2019-07-31
Support for no Instant::now()

### LINKED ISSUE #166: Use of Instant::now() does not work on `wasm32-unknown-unknown`
The underlying sys_call is not implemented for `Instant::now()` for the `wasm32-unknown-unknown` target (unless recompiling rust with `xargo` and enabling `wasm_syscall`):

```
panicked at 'Time system call is not implemented by WebAssembly host', src/libstd/sys/wasm/mod.rs:302:13

Stack:

Error
    at Module.__wbg_new_59cb74e423758ede (webpack:///../pkg/spawn_chain.js?:181:26)
    at __wbg_new_59cb74e423758ede (http://localhost:8080/bootstrap.js:65:102)
    at console_error_panic_hook::hook::h8df71f3722ab18fc (wasm-function[194]:292)
    at core::ops::function::Fn::call::h42e98a3c026ddf0b (wasm-function[776]:3)
    at std::panicking::rust_panic_with_hook::h3f94f83752aaaaa5 (wasm-function[314]:265)
    at std::panicking::begin_panic::h6beec07bc67d7532 (wasm-function[613]:40)
    at std::sys::wasm::TimeSysCall::perform::h27f1627f17fac1d9 (wasm-function[721]:13)
    at std::sys::wasm::time::Instant::now::h34de02a8c4aa47a2 (wasm-function[770]:3)
    at std::time::Instant::now::h70048c3feeca6e27 (wasm-function[791]:1)
    at parking_lot_core::parking_lot::HashTable::new::h1383e3e985aa35d8 (wasm-function[208]:53)
```

Finding a way to substitute the use of `Instant` in configuration or using a hooks system would be helpful to make this work for `wasm32-unknown-unknown` to allow plugging in a time source that works for Web vs WASI, etc.

KronicDeth:
Fixes #166

For `wasm32-unknown-unknown` (and I think SGX) `Instant::now()` is not
supported because it cannot assume a time source as reported in #166.

To allow for substituting `Instant::now()` at runtime as is needed for `wasm32-unknown-unknown` targeting Web browsers that would be using `performance.now()`, have all code that hard-coded `std::time::Instant`, `std::time::Instant::now()`, and `std::time::Instant::checked_duration_since` use types and functions in `parking_lot_core::time`.

## Without `no_instant_now` feature

* `parking_lot_core::time::Instant` is `std::time::Instant`
* `parking_lot_core::time::now()` is `std::time::Instant::now()`
* `parking_lot_core::time::checked_duration_since(later, earlier)` is `std::time::Instant::checked_duration_since`

## With `no_instant_now` feature

* `parking_lot_core::time::Instant` is `std::time::Duration`.  (This was chosen as the rust `sys` for `wasm` uses `Duration` as the internal field type for `Instant`s, so it seemed safe to depend on it.) . `Instant` can't be used because it is opaque and not safely fakable when `Instant::now` can't be used because `Instant` internals vary by platform.
* `parking_lot_core::time::now()` will panic by default. `parking_lot_core::time::set_now_fn(now_fn)` must be called with a `fn() -> Instant`, such as one that defers to `performance.now()` for browsers and used `Duration::new` or one of the `from` functions to make a `Duration` that can be returned from the `Instant`.
* `parking_lot_coer::time::checked_duration_since(later, earlier)` is `std::time::Duration::checked_sub`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ owning_ref = ["lock_api/owning_ref"]
 nightly = ["parking_lot_core/nightly", "lock_api/nightly"]
 deadlock_detection = ["parking_lot_core/deadlock_detection"]
 serde = ["lock_api/serde"]
+no_instant_now = ["parking_lot_core/no_instant_now"]
 
 [workspace]
 exclude = ["benchmark"]
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -34,4 +34,5 @@ winapi = { version = "0.3", features = ["winnt", "ntstatus", "minwindef",
 
 [features]
 nightly = []
+no_instant_now = []
 deadlock_detection = ["petgraph", "thread-id", "backtrace"]
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -59,6 +59,7 @@
 mod parking_lot;
 mod spinwait;
 mod thread_parker;
+pub mod time;
 mod util;
 mod word_lock;
 
--- a/core/src/parking_lot.rs
+++ b/core/src/parking_lot.rs
@@ -6,6 +6,7 @@
 // copied, modified, or distributed except according to those terms.
 
 use crate::thread_parker::{ThreadParker, ThreadParkerT, UnparkHandleT};
+use crate::time::{self, Instant};
 use crate::util::UncheckedOptionExt;
 use crate::word_lock::WordLock;
 use core::{
@@ -14,7 +15,7 @@ use core::{
     sync::atomic::{AtomicPtr, AtomicUsize, Ordering},
 };
 use smallvec::SmallVec;
-use std::time::{Duration, Instant};
+use std::time::Duration;
 
 static NUM_THREADS: AtomicUsize = AtomicUsize::new(0);
 static HASHTABLE: AtomicPtr<HashTable> = AtomicPtr::new(ptr::null_mut());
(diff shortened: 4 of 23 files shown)


CryZe:
The feature should be called `has_instant_now` or something. Features are additive, not subtractive.

KronicDeth:
@CryZe I was thinking `no` like `nostd` because `wasm32-unknown-unknown` is missing `Instant::now()` support (it exists, but panics).  I'm fine with switching the logic though if the team wants it.

CryZe:
`no_std` is not a valid feature either. All the crates use `std` as the feature because the opposite isn't properly supported by cargo.

Amanieu:
I'm not to familiar with WASM, so I'll ping @alexcrichton for his opinion.

Amanieu:
I'm not too familiar with WASM so I'll ping @alexcrichton for his opinion.

alexcrichton:
It's true that wasm doesn't have any native support for `Instant`. I think there's unfortunately not a great answer for how to support `parking_lot` in wasm for now. The current status quo, which isn't great, is to have features on the crate which select how a crate works on wasm. For example by default this crate would panic if `Instant::now()` is needed on wasm. The `wasm-bindgen` feature, perhaps, though would enable usage of `js-sys` and the `Date` type in JS. Perhaps a future runtime would have its own support if WebAssembly is being run outside of the browser.

This isn't really a scalable workable solution in the long run but it's working "well enough" for now. In that sense I'd recommend doing that rather than providing a hook to provide the ability to get an `Instant`.

KronicDeth:
From @alexcrichton suggestion, I'm going to make this less general and instead use these criteria:

1. `target_arch` is `wasm32`
2. `feature` `wasm-bindgen` is enabled

Then use `web_sys` to use `performance.now()` as the clock source directly.  I'll make a new PR when it is ready.

cormacrelf:
@KronicDeth how did you go with this? I have a WebAssembly crate that depends on Salsa, which uses parking_lot under the hood. The solution in [your commit here](https://github.com/KronicDeth/parking_lot/commit/9866be37213976053a1272e809410a62ad68a803) seems perfect -- could you craft a PR here so others can use it more easily?

PR #170, opened 2019-09-19
Switch from `rustc_version` to `autocfg`

alexcrichton:
The `autocfg` crate is a bit more robust in terms of how it parses
rustc's version and also doesn't have a dependency on `semver` which can
be a very weighty dependency if its `serde` feature is activated.


Amanieu:
bors retry

PR #172, opened 2019-09-23
Use full path to lock_api traits in other crates

faern:
My main pet peeve here was the renaming of a trait during import because it collided with an existing type. The main problem here IMO was `impl RawMutexTrait for RawMutex`. Then one has to remember that `RawMutexTrait` corresponds to `lock_api::RawMutex`. This is yet another step to take in the head when reading the code. Just using `lock_api::RawMutex` is not much longer, and it immediately makes it much more clear to the reader that the *local type* `RawMutex` implements the trait *from another crate*. Clears up the relationship with these crates a little bit.

For consistency, and for parts of my argument above, I opted to implement all traits in the same files with full paths to the traits, not just the colliding one.

If/when we drop support for Rust 1.31 and 1.32 we can change the imports to just `use lock_api::RawMutex as _` to make it even better. No name collision and no need to invent yet another name. Until then I opted for this new name that just tries to avoid the collision in the simplest way possible without adding a word that should not be there.

### DIFF
--- a/src/condvar.rs
+++ b/src/condvar.rs
@@ -12,7 +12,7 @@ use core::{
     fmt, ptr,
     sync::atomic::{AtomicPtr, Ordering},
 };
-use lock_api::RawMutex as RawMutexTrait;
+use lock_api::RawMutex as RawMutex_;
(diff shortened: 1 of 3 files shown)


Amanieu:
bors retry

PR #173, opened 2019-09-23
Remove some usage of deprecated mem::uninitialized

faern:
`mem::uninitialized` was deprecated in 1.38. Usage is not recommended. Here I start getting rid of it in favor of the preferred `mem::MaybeUninit` type.

This PR does not yet replace all occurances of `mem::unintialized`. I wanted to get some feedback on this before I went ahead with the CloudABI thread parker as well.

### DIFF
--- a/src/remutex.rs
+++ b/src/remutex.rs
@@ -17,8 +17,9 @@ unsafe impl GetThreadId for RawThreadId {
 
     fn nonzero_thread_id(&self) -> NonZeroUsize {
         // The address of a thread-local variable is guaranteed to be unique to the
-        // current thread, and is also guaranteed to be non-zero.
-        thread_local!(static KEY: u8 = unsafe { ::std::mem::uninitialized() });
+        // current thread, and is also guaranteed to be non-zero. The variable has to have a
+        // non-zero size to guarantee it has a unique address for each thread.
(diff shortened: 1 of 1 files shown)


faern:
Ok. This requires Rust 1.36. I forgot to check that.

faern:
So I reverted all the changes that started using `MaybeUninit`. But there was one place where I had simply removed `mem::uninitialized`, so maybe that is still viable to merge already.

I'm not sure if the `KEY` has to occupy space or why it could not simply be initialized to `0`. I assume the latter was to let the compiler maybe optimize a bit more since the value was never read anyway. I tried replacing it with a zero sized unit type. Should not cost any memory, nor actual initialization. The question is if the compiler will actually make the code generate sane pointers for `KEY`, but the tests passed at least.

Amanieu:
We actually *do* want `KEY` to occupy at least 1 byte of memory so that it is guaranteed to have a unique address for each thread.

Amanieu:
But initializing it with 0 instead of leaving it uninitialized is probably fine.

Amanieu:
bors retry

Amanieu:
Travis issue on OSX, so I'm just going to merge manually.

Amanieu:
bors retry

PR #174, opened 2019-09-25
Add documentation and limit number of expressions in unsafe block for RawMutex

faern:
Try to move not unsafe stuff out of large unsafe blocks. And keep only the unsafe call at the end in that block. Documenting the safety contract to think about in said call.

Document what the bits in `RawMutex::state`, and what all combinations of that state, means. Please verify the descriptions are correct, since you know all edge cases much better than me :)

Clippy complained about `#[inline]` on function with no body. The corresponding actual implementations still have `#[inline]`

Clippy complained about two unsafe functions with docs, but without a `# Safety` section. So I took what looked like the safety notes of those functions and formatted them according to the standard.

### DIFF
--- a/src/raw_mutex.rs
+++ b/src/raw_mutex.rs
@@ -220,37 +220,41 @@ impl RawMutex {
             }
 
             // Park our thread until we are woken up by an unlock
-            unsafe {
-                let addr = self as *const _ as usize;
-                let validate = || self.state.load(Ordering::Relaxed) == LOCKED_BIT | PARKED_BIT;
-                let before_sleep = || {};
-                let timed_out = |_, was_last_thread| {
-                    // Clear the parked bit if we were the last parked thread
-                    if was_last_thread {
-                        self.state.fetch_and(!PARKED_BIT, Ordering::Relaxed);
-                    }
-                };
-                match parking_lot_core::park(
+            let addr = self as *const _ as usize;
+            let validate = || self.state.load(Ordering::Relaxed) == LOCKED_BIT | PARKED_BIT;
+            let before_sleep = || {};
+            let timed_out = |_, was_last_thread| {
+                // Clear the parked bit if we were the last parked thread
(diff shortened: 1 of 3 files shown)


src/raw_mutex.rs
```diff
@@ -28,11 +28,30 @@ pub(crate) const TOKEN_NORMAL: UnparkToken = UnparkToken(0);
...
+    ///     0      |     0      | The mutex is not locked, nor is anyone waiting for it.
+    /// -----------+------------+------------------------------------------------------------------
+    ///     0      |     1      | The mutex is locked by exactly one thread. No other thread is
+    ///            |            | waiting for it.
+    /// -----------+------------+------------------------------------------------------------------
+    ///     1      |     0      | The mutex is not locked. There are parked threads waiting for it.
+    ///            |            | At least one of the parked threads are just about to be unparked.
+    /// -----------+------------+------------------------------------------------------------------
+    ///     1      |     1      | The mutex is locked by exactly one thread. At least one thread is
+    ///            |            | parked, waiting for the lock to become available.
```
Amanieu: I would add an additional comment here because this is actually quite subtle:
- `PARKED_BIT` is only ever cleared when a bucket lock is held (i.e. in a `parking_lot_core` callback). This ensures that we never end up in a situation where there are parked threads but `PARKED_BIT` is not set (which would result in those threads potentially never getting woken up).
- We set `PARKED_BIT` before parking, so technically `PARKED_BIT` means "one or more thread is parked or about to park`. The `validate` callback is very important here since we need to ensure that `PARKED_BIT` is still set after we acquire the bucket lock (since a concurrent unpark may have cleared it).

faern: Great additions and clarifications. Copied some of it verbatim.

Amanieu:
Travis issue on OSX, so I'm just going to merge manually.

PR #179, opened 2019-09-27
Add unsafe safety notes to raw_rwlock.rs

faern:
Reducing the amount of expressions inside `unsafe` blocks. And adding `// SAFETY:` documentation to all remaining unsafe blocks.

Found two methods where the unsafety leaked out of methods not marked as unsafe(!). So I made them unsafe and documented what has to be upheld.

Found potential panic in closure that is not allowed to panic. But only in debug builds. What do we do about that?

### DIFF
--- a/src/raw_rwlock.rs
+++ b/src/raw_rwlock.rs
@@ -713,16 +713,15 @@ impl RawRwLock {
         // just need to wake up a potentially sleeping pending writer.
         // Using the 2nd key at addr + 1
         let addr = self as *const _ as usize + 1;
-        let callback = |result: UnparkResult| {
+        let callback = |_result: UnparkResult| {
             // Clear the WRITER_PARKED_BIT here since there can only be one
             // parked writer thread.
-            debug_assert!(!result.have_more_threads);
             self.state.fetch_and(!WRITER_PARKED_BIT, Ordering::Relaxed);
             TOKEN_NORMAL
         };
         // SAFETY:
         //   * `addr` is an address we control.
-        //   * `callback` does not panic (in non debug) or call into any function of `parking_lot`.
+        //   * `callback` does not panic or call into any function of `parking_lot`.
         unsafe {
             parking_lot_core::unpark_one(addr, callback);
         }
--- a/src/raw_rwlock.rs
+++ b/src/raw_rwlock.rs
@@ -1085,40 +1085,39 @@ impl RawRwLock {
             }
 
             // Park our thread until we are woken up by an unlock
-            unsafe {
-                let addr = self as *const _ as usize;
-                let validate = || {
-                    let state = self.state.load(Ordering::Relaxed);
-                    state & PARKED_BIT != 0 && validate(state)
-                };
-                let before_sleep = || {};
-                let timed_out = |_, was_last_thread| {
-                    // Clear the parked bit if we were the last parked thread
-                    if was_last_thread {
-                        self.state.fetch_and(!PARKED_BIT, Ordering::Relaxed);
-                    }
-                };
-                match parking_lot_core::park(
-                    addr,
-                    validate,
-                    before_sleep,
-                    timed_out,
-                    token,
-                    timeout,
-                ) {
-                    // The thread that unparked us passed the lock on to us
-                    // directly without unlocking it.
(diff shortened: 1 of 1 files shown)


src/raw_rwlock.rs
```diff
@@ -652,114 +652,118 @@ impl RawRwLock {
...
     #[cold]
     fn unlock_shared_slow(&self) {
         // At this point WRITER_PARKED_BIT is set and READER_MASK is empty. We
         // just need to wake up a potentially sleeping pending writer.
+        // Using the 2nd key at addr + 1
+        let addr = self as *const _ as usize + 1;
+        let callback = |result: UnparkResult| {
+            // Clear the WRITER_PARKED_BIT here since there can only be one
+            // parked writer thread.
+            debug_assert!(!result.have_more_threads);
```
faern: Seems like for debug builds, we don't uphold the safety requirement to never panic in `callback`. This seems bad. We don't want UB in debug builds.

faern: If we really want to make this sanity check here, and we can't panic, maybe we need to resort to aborting the process.

Amanieu: Well this assert never triggers in practice, so it should be fine. These kinds of debug asserts are for debugging issues in parking_lot itself rather than for anything depending on user input.

src/raw_rwlock.rs
```diff
@@ -898,41 +911,49 @@ impl RawRwLock {
...
+    /// `parking_lot`.
     #[inline]
-    fn wake_parked_threads<C>(&self, new_state: usize, callback: C)
-    where
-        C: FnOnce(usize, UnparkResult) -> UnparkToken,
-    {
+    unsafe fn wake_parked_threads(
+        &self,
+        new_state: usize,
+        callback: impl FnOnce(usize, UnparkResult) -> UnparkToken,
```
faern: Not related to the safety. Went ahead and made this, and two more places, `impl Trait`s instead. They are just internal functions so should not affect the API.

faern:
Oh. Wait.. All the `validate` closures sent to `lock_common` have a very similar pattern. All of them do `|state| state & some_value != 0`. We can just modify `lock_common` to take `some_value` and define the closure locally. Then we can know it upholds the safety properties.

faern:
No test case seems to cover the `validate` part of `lock_common`. Because I can intentionally screw up the logic and all tests still pass. That is not very reassuring. But I think I implemented it correctly.

src/raw_rwlock.rs
faern: Won't the assert they be compiled into any project depending on `parking_lot` that does not build with `--release`? So it's present for anyone debugging anything, not just `parking_lot` itself? 🤔

Amanieu: But it doesn't matter since that code is literally unreachable. I only use debug asserts as a way to document invariants, not because I actually expect them to panic.

faern: If I depend on `parking_lot` in my crate and build a debug build. I'm still counting on it being being 100% memory safe.

faern: I wrote that last comment before my browser had loaded your previous comment..

So. By "literally unreachable" you mean the assert will only panic if `parking_lot` has a bug? Yeah, sure. I don't know. Still does not uphold the safety requirements, if the safety requirements are correctly stated that is.

faern: How about creating a feature (that is disabled by default) and an internal version of `debug_assert!` that does nothing when the feature is disabled? Then the documentational purposes of these asserts remain the same, but we stop breaking guarantees.

Amanieu: Just remove the assert then if you feel that strongly about it.

faern: Done.
But then I started thinking. Is it really unsafe to panic in this `callback`? if `unpark_one` is aborted at the point where it calls `callback` it will leave the bucket for this key locked forever. That will of course cause a deadlock. But deadlocks are not unsafe. Can something else happen if it panics that leads to UB?

Amanieu:
It's a bit difficult to see from the diff, but I assume you didn't modify the contents of any of the closures, only moved them to separate variables?

faern:
Exactly. So they would be defined outside of the unsafe function call.

PR #185, opened 2019-10-03
Replace deprecated mem::uninitialized()

faern:
Follow up to #173. If we vendor the parts of `MaybeUninit` we need for the places where we use it, then we can get rid of all `mem::uninitialized` and still support Rust older than 1.36.

Because my impression is that there is no way `parking_lot` will make it into `std` while using methods `std` has deprecated and explicitly stated no one should use.

I was not able to make the vendored version `#[repr(transparent)]`.

### DIFF
--- a/core/src/maybe_uninit.rs
+++ b/core/src/maybe_uninit.rs
@@ -19,22 +19,22 @@ pub union MaybeUninit<T: Copy> {
 
 #[cfg(not(has_maybe_uninit))]
 impl<T: Copy> MaybeUninit<T> {
-    #[inline(always)]
+    #[inline]
     pub fn uninit() -> MaybeUninit<T> {
         MaybeUninit { uninit: () }
     }
 
-    #[inline(always)]
+    #[inline]
     pub fn as_ptr(&self) -> *const T {
         unsafe { &*self.value as *const T }
     }
 
-    #[inline(always)]
+    #[inline]
     pub fn as_mut_ptr(&mut self) -> *mut T {
         unsafe { &mut *self.value as *mut T }
     }
 
-    #[inline(always)]
+    #[inline]
     pub unsafe fn assume_init(self) -> T {
         ManuallyDrop::into_inner(self.value)
     }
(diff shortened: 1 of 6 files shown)


core/src/maybe_uninit.rs
```diff
@@ -0,0 +1,41 @@
...
+}
+
+#[cfg(not(has_maybe_uninit))]
+impl<T: Copy> MaybeUninit<T> {
+    #[inline(always)]
+    pub fn uninit() -> MaybeUninit<T> {
+        MaybeUninit { uninit: () }
+    }
+
+    #[inline(always)]
```
Amanieu: Is `#[inline(always)]` really necessary here? Would plain `#[inline]` be sufficient?

faern: I can change it. I just copied the implementation and annotations from `libcore`

faern: I have not looked at any generated code to see whether it makes a difference.

Amanieu: I would prefer changing it to a simple `#[inline]`.

faern: Ok. I personally don't really see why we would want to write the copied version differently than the original. But I can push that! Do you mean for this method in particular or everything on our `MaybeUninit`?

Amanieu: For all of the methods.

faern: Done

faern:
As I wrote, `#[repr(transparent)]` does not work on `union`s. But we could add `#[repr(C)]`, that works. Do you think it's worth it? I'm not aware how Rust lays out `union`s without explicit `repr` specified.

Amanieu:
Considering pre-1.36 compilers aren't being updated anymore, I think it's safe to rely on undocumented behavior (struct layout for single-field structs) for those.

faern:
Looks like travis tries to use the branch `staging.tmp` while in reality it's just called `staging`

Amanieu:
bors retry

PR #186, opened 2019-10-09
Safe grow_hashtable

faern:
Calling `grow_hashtable` can't do unsafe stuff unless `HASHTABLE` has been wrongfully manipulated. So there are no safety invariants to uphold as a caller. So it does not need to be `unsafe`. The little unsafe Rust used inside it has instead been contained and documented.

Moving the hardest to understand/read part, linked list handling, into a separate function for isolation and to make it easier to understand on its own.


Amanieu:
bors retry

PR #188, opened 2019-11-04
Bump smallvec version to 1.0

messense:
See https://github.com/servo/rust-smallvec/pull/175 for changelog


Amanieu:
CI is failing, you'll need to bump the minimum Rust version.

PR #189, opened 2019-11-05
Remove pre 1.36 backwards compatibility code

faern:
Back when I submitted #185 I did not get the impression that dropping support for older Rust was an option. But it seemed to have gone through without much fanfare in #188, so I guess we can remove all this backwards compatibility code now then.


Amanieu:
Yea I merged that because I feel that we're going to be publishing a new major version soon anyways.

PR #191, opened 2019-11-16
Add initial parking_lot_core tests

faern:
This library could use more tests, to be more robust. This was one of the things pointed out in the review of the integration into std in https://github.com/rust-lang/rust/pull/56410

`WTF::ParkingLot` has a number of tests we can take inspiration from. And this is somewhat of a port of some of those tests. However, when I ported it 1:1 I found a race condition, so I had to implement the semaphore a bit differently.

I also got away without the tests relying on a working mutex or condvar implementation. Because we can't make the `parking_lot_core` tests depend on the higher level `parking_lot`, nor do we want to depend on locking primitives in std if we aim to become the base for those implementations one day.

### DIFF
--- a/core/src/parking_lot.rs
+++ b/core/src/parking_lot.rs
@@ -1439,18 +1439,42 @@ mod tests {
     }
 
     test! {
-        unpark_all_one_fast(repeats: 10000, latches: 1, delay: 0, threads: 1, single_unparks: 0);
-        unpark_all_hundred_fast(repeats: 100, latches: 1, delay: 0, threads: 100, single_unparks: 0);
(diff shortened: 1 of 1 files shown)


Amanieu:
LGTM. Do you think it would be possible to extend these to cover `unpark_filter` and `unpark_requeue`? Or would this be better done as a separate set of tests?

faern:
I'm aiming for adding more tests. So we at least touch the entire public API of `parking_lot_core` at least once. Not sure if those would use the semaphore, probably not(?)
And I felt I should add what `WTF::ParkingLot` was testing first, and then extend once we have agreed this type of testing is sane.

PR #192, opened 2019-11-22
Add integration tests for Condvar

pinkisemils:
As discussed during Impl days, I've gone and pretty much verbatim copied WebKit's `Condvar` tests here.

### DIFF
--- a/src/condvar.rs
+++ b/src/condvar.rs
@@ -693,8 +693,8 @@ mod tests {
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/condvar.rs
```diff
@@ -692,3 +692,392 @@ mod tests {
         }
     }
 }
+
+/// This module contains an integration test that is verbatim stolen from WebKit's Condvar
+/// integration test. It's a thread-safe Queue that should
```
Amanieu: You forgot to finish your sentence here...

PR #193, opened 2019-11-24
Misc clippy fixes

faern:
I went over the crates with Clippy. Some suggested changes are only stylistic, others more optimization oriented. The commit messages should hopefully explain well enough why each change was made. Otherwise see what clippy says on current master :)

I left some warnings, since the suggested changes were not really objectively better, or possibly larger than what I wanted to do right now.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,9 +21,6 @@ lazy_static = "1.0"
 # Used when testing out serde support.
 bincode = {version = "1.1.3"}
 
-[build-dependencies]
-autocfg = "0.1.6"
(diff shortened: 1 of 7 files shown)


faern:
Also pulled in the removal of the `autocfg` dependency that we stopped using over in #189

Amanieu:
This is technically a breaking change, but we're about to make a major version bump anyways...

faern:
My thought exactly. I'm bumping versions and drafting a changelog at the moment. Will submit soon.

PR #194, opened 2019-11-24
Add changelog entries and bump versions for 0.10.0 release

faern:
We must first let #193 be merged. Otherwise this changelog is not even correct.

There were a lot of changes since last release. But the vast majority is reformatting and re-arranging to reduce the number of unsafe expressions. As well as adding safety invariant documentation and a few tests. These are cool improvements, but does not affect the public API. I can add something about it in the changelog if you think it's worth it.

`lock_api` only has formatting changes and a few documentation improvements, so I don't think it needs a breaking version change. But might be worth releasing anyway to get the improved documentation out.

The date in the changelog is simply tomorrow. Since I did not think we would have time to release today, and I did not want to just go with `2019-XX-YY` either. If we don't think we'll release tomorrow we can just fix that before merging this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## parking_lot 0.10.0, parking_lot_core 0.7.0, lock_api 0.3.2 (2019-11-25)
+
+- Upgrade smallvec dependency to 1.0 in parking_lot_core.
+- Replace all usage of `mem::unitialized` with `mem::MaybeUninit`.
+- The minimum required Rust version is bumped to 1.36. Because of the above two changes.
+- Make methods on `WaitTimeoutResult` and `OnceState` take `self` by value instead of reference.
+
 ## parking_lot_core 0.6.2 (2019-07-22)
 
 - Fixed compile error on Windows with old cfg_if version. (#164)
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "parking_lot"
-version = "0.9.0"
+version = "0.10.0"
 authors = ["Amanieu d'Antras <amanieu@gmail.com>"]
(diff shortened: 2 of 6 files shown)


faern:
Except for the tiny `&self -> self` change in #193 the only breaking change I can see is the MSRV bump.

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## parking_lot 0.10.0, parking_lot_core 0.70, lock_api 0.3.2 (2019-11-25)
```
Amanieu: Should be 0.7.0.

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
...
+
+- Upgrade smallvec dependency to 1.0 in core.
+- Replace all usage of mem::unitialized with mem::MaybeUninit.
```
Amanieu: Use backticks for code.

Cargo.toml
```diff
@@ -11,7 +11,7 @@ categories = ["concurrency"]
 edition = "2018"
 
 [dependencies]
-parking_lot_core = { path = "core", version = "0.6.2" }
+parking_lot_core = { path = "core", version = "0.7.0" }
 lock_api = { path = "lock_api", version = "0.3.1" }
```
faern: I'm not bumping the dependency on `lock_api` here. Because the new `parking_lot` does not *need* the new release. It's just some documentation fixes.

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## parking_lot 0.10.0, parking_lot_core 0.70, lock_api 0.3.2 (2019-11-25)
+
+- Upgrade smallvec dependency to 1.0 in core.
```
Amanieu: s/core/parking_lot_core/ to avoid confusion with libcore.

PR #195, opened 2019-11-25
Remove rust2015 style docs

faern:
We juuuust had a release. But when sifting through the docs I stumbled upon a `extern crate parking_lot;`. That felt very ancient. So I removed it.

I also found an example with `use` statements that were not sorted. So very minor, but could fix it while I was at it anyway.


Amanieu:
bors retry

PR #203, opened 2020-01-06
fix: not panic in destructor

zhangli-pear:
`release_resource` should not panic. Because `with_thread_data` may return a `ThreadData` on the stack, and it's empty.

### DIFF
--- a/core/src/parking_lot.rs
+++ b/core/src/parking_lot.rs
@@ -1198,10 +1198,9 @@ mod deadlock_impl {
     pub unsafe fn release_resource(key: usize) {
         with_thread_data(|thread_data| {
             let resources = &mut (*thread_data.deadlock_data.resources.get());
-            match resources.iter().rposition(|x| *x == key) {
-                Some(p) => resources.swap_remove(p),
-                None => panic!("key {} not found in thread resources", key),
-            };
+            if let Some(p) = resources.iter().rposition(|x| *x == key) {
+                resources.swap_remove(p);
+            }
         });
     }
 


zhangli-pear:
following code can reproduce the problem.

```rust
use std::thread;
use parking_lot::RwLock;

struct Bar(RwLock<()>);

impl Drop for Bar {
    fn drop(&mut self) {
        let _n = self.0.write();
    }
}

thread_local! {
    static B: Bar = Bar(RwLock::new(()));
}

fn main() {
    thread::spawn(|| {
        let a = RwLock::new(());
        let _a = a.read();

        B.with(|_| ());
    }).join().unwrap();
}
```

Amanieu:
I can't reproduce the panic with your example code. Are you sure you are using the latest version of parking_lot?

zhangli-pear:
> I can't reproduce the panic with your example code. Are you sure you are using the latest version of parking_lot?

Did you enable `deadlock_detection` feature? 
I'm using 0.9.0 in my project, but latest version is the same.

Amanieu:
Yes I enabled `deadlock_detection`. However no matter how I run your code, I can't get it to panic.

Amanieu:
Can you try deleting your Cargo.lock to see if this helps?

quininer:
This seems to happen only on macOS, it is same issue as https://github.com/Amanieu/parking_lot/issues/114.

Amanieu:
Right I think I know what the problem is. Basically, the internal `ThreadData` in parking_lot is getting dropped before the `RwLock`, which causes all sorts of issues. I'll need to re-read the code to figure out all the implications.

quininer:
```rust
use std::thread;
use parking_lot::RwLock;

struct Bar(RwLock<()>);

impl Drop for Bar {
    fn drop(&mut self) {
        let _n = self.0.write();
    }
}

thread_local! {
    static B: Bar = Bar(RwLock::new(()));
}

fn main() {
    thread::spawn(|| {
        B.with(|_| ());

        let a = RwLock::new(());
        let _a = a.read();
    }).join().unwrap();
}
```

reversing init order of tld can make it reproduce on linux.

Amanieu:
I think a more general solution is needed here: the deadlock detector needs to be able to handle threads that have exited while still holding leaked locks.

zhangli-pear:
> I think a more general solution is needed here: the deadlock detector needs to be able to handle threads that have exited while still holding leaked locks.

maybe not using thread_local?

Amanieu:
Actually after reviewing the code, I'm just going to go with your solution.

PR #204, opened 2020-01-06
Create API for mutex that is always fair

### LINKED ISSUE #76: Consider making "fairness" something decided when a lock is created
Currently we have `MutexGuard::unlock_fair()`. It'd be quite convenient for me to instead have something like that:

```rust
let lock = Mutex::new_fair(()); 
let lock = Mutex::new((), /* fair: */ false);  // or
```

Which would allow the normal `Drop` to be fair or unfair.

paulocsanz:
Fixes #76. I don't know if there is interest in having a API for a Mutex that is always fair, but solving it at the type level seems better than requiring a state for each mutex. Fairness sometimes is required, so making type level assurances so another dev can't screw up may solve a bunch of problems, it's probably already being done by people that need it since it's just a thin wrapper. 

This implements a `FairMutex` and it's `RawFairMutex`. This could be done for RwLock too if there is interest.

Idk if the tests are required since it's literally calling something already tested, but I just copied the files and made it compile. I had to remove the Condvar test, since Condvar assumes it's a MutexGuard, and doesn't accept a FairMutexGuard.

### DIFF
--- a/src/fair_mutex.rs
+++ b/src/fair_mutex.rs
@@ -86,7 +86,7 @@ pub type MappedFairMutexGuard<'a, T> = lock_api::MappedMutexGuard<'a, RawFairMut
 
 #[cfg(test)]
 mod tests {
-    use crate::{Condvar, FairMutex};
+    use crate::FairMutex;
     use std::sync::atomic::{AtomicUsize, Ordering};
     use std::sync::mpsc::channel;
     use std::sync::Arc;
@@ -95,13 +95,8 @@ mod tests {
     #[cfg(feature = "serde")]
     use bincode::{deserialize, serialize};
 
-    struct Packet<T>(Arc<(FairMutex<T>, Condvar)>);
-
     #[derive(Eq, PartialEq, Debug)]
     struct NonCopy(i32);
-
-    unsafe impl<T: Send> Send for Packet<T> {}
-    unsafe impl<T> Sync for Packet<T> {}
 
     #[test]
     fn smoke() {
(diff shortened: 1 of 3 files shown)


src/fair_mutex.rs
```diff
@@ -0,0 +1,259 @@
...
+/// - No poisoning, the lock is released normally on panic.
+/// - Only requires 1 byte of space, whereas the standard library boxes the
+///   `FairMutex` due to platform limitations.
+/// - Can be statically constructed (requires the `const_fn` nightly feature).
+/// - Does not require any drop glue when dropped.
+/// - Inline fast path for the uncontended case.
+/// - Efficient handling of micro-contention using adaptive spinning.
+/// - Allows raw locking & unlocking without a guard.
+/// - Supports eventual fairness so that the mutex is fair on average.
+/// - Optionally allows making the mutex fair by calling `FairMutexGuard::unlock_fair`.
```
Amanieu: These last 2 sentences don't apply to `FairMutex`.

src/fair_mutex.rs
```diff
@@ -0,0 +1,259 @@
...
+/// This mutex will block threads waiting for the lock to become available. The
+/// mutex can also be statically initialized or created via a `new`
+/// constructor. Each mutex has a type parameter which represents the data that
+/// it is protecting. The data can only be accessed through the RAII guards
+/// returned from `lock` and `try_lock`, which guarantees that the data is only
+/// ever accessed when the mutex is locked.
+///
+/// This mutex is always fair.
+///
+/// # Differences from the standard library `FairMutex`
```
Amanieu: The standard library doesn't have a `FairMutex`. This should just be `Mutex`.

src/fair_mutex.rs
```diff
@@ -0,0 +1,259 @@
...
+/// A always fair mutual exclusion primitive useful for protecting shared data
+///
+/// This mutex will block threads waiting for the lock to become available. The
+/// mutex can also be statically initialized or created via a `new`
+/// constructor. Each mutex has a type parameter which represents the data that
+/// it is protecting. The data can only be accessed through the RAII guards
+/// returned from `lock` and `try_lock`, which guarantees that the data is only
+/// ever accessed when the mutex is locked.
+///
+/// This mutex is always fair.
```
Amanieu: This sentence is redundant, this is already stated in the summary at the top. Instead, an explanation of what fairness is exactly would be more useful. This should explain why a fair mutex is slower than an unfair one, and when fairness is useful (latency vs throughput).

paulocsanz: Is the description good now?

Issue #206, opened 2020-01-12
Ambiguous Documentation for `try_lock`


Tankobot:
I'm currently using the `lock_api` crate for my own `no_std` crate and I think the documentation for `RawMutex::try_lock` is missing information. It is [stated](https://docs.rs/lock_api/0.3.3/lock_api/trait.RawMutex.html#required-methods) that a `bool` will be returned, but nothing is said about what the `bool` indicates. If we look at the [example](https://docs.rs/lock_api/0.3.3/lock_api/#example) given at the top of the crate, we can infer that `true` should be returned on failure since swapping a `true` atomic bool will return `true`. In any case, clearer documentation would be nice and I am uncertain if I have perhaps misinterpreted the example implementation.

Amanieu:
Oops, the example is wrong. In fact `try_lock` should return `true` only if it succeeded in acquiring the lock.

Tankobot:
Would you like me to submit a pull request to fix the example and add a line to the `RawMutex::try_lock` documentation?

Amanieu:
Yes, please do.

Amanieu:
Fixed by #207

PR #207, opened 2020-01-13
Fix documentation and example for `try_lock`

Tankobot:
As described in #206:

The example implementation for `try_lock` from the `lock_api` crate was incorrect and has been fixed. Additionally a description of what the `bool` returned from `try_lock` means has been added.

### DIFF
--- a/lock_api/src/lib.rs
+++ b/lock_api/src/lib.rs
@@ -47,7 +47,9 @@
 //!     }
 //!
 //!     fn try_lock(&self) -> bool {
-//!         self.0.swap(true, Ordering::Acquire)
+//!         self.0
+//!             .compare_exchange(false, true, Ordering::Acquire, Ordering::Relaxed)
+//!             .is_ok()
 //!     }
 //!
 //!     fn unlock(&self) {
--- a/lock_api/src/mutex.rs
+++ b/lock_api/src/mutex.rs
@@ -40,7 +40,8 @@ pub unsafe trait RawMutex {
     /// Acquires this mutex, blocking the current thread until it is able to do so.
     fn lock(&self);
 
-    /// Attempts to acquire this mutex without blocking.
+    /// Attempts to acquire this mutex without blocking. Returns `true`
+    /// if the lock was successfully acquired and `false` otherwise.
     fn try_lock(&self) -> bool;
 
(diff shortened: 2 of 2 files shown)


lock_api/src/lib.rs
```diff
@@ -47,7 +47,13 @@
...
 //!
 //!     fn try_lock(&self) -> bool {
-//!         self.0.swap(true, Ordering::Acquire)
+//!         match self
+//!             .0
+//!             .compare_exchange_weak(false, true, Ordering::Acquire, Ordering::Relaxed)
+//!         {
+//!             Ok(_) => true,
+//!             Err(_) => false,
+//!         }
```
Amanieu: You want a strong cmpxchg for `try_lock`, to avoid spurious failures. Also just use `.is_ok()` instead of a match.

Tankobot: Curious, why is a strong cmpxchg necessary? Does it matter if the call fails? The documentation merely says that acquisition is attempted, not that it is guaranteed if the lock is available.

lock_api/src/lib.rs
```diff
@@ -47,7 +47,10 @@
 //!     }
 //!
 //!     fn try_lock(&self) -> bool {
-//!         self.0.swap(true, Ordering::Acquire)
+//!         match self
+//!             .0
+//!             .compare_exchange(false, true, Ordering::Acquire, Ordering::Relaxed)
+//!             .is_ok()
```
Tankobot: Changed to strong cmpxchg and `is_ok` method.

lock_api/src/lib.rs
Tankobot: I pushed the changes.

Amanieu:
You seem to have picked up some extra commits in your PR

Amanieu:
Because on some platforms (e.g. ARM) a weak cmpxchg can randomly fail for arbitrary reasons (such as a timer interrupt), even if the lock is currently free. Weak cmpxchg is generally used in a loop where you keep retrying until the cmpxchg succeeds, which isn't the case here.

Amanieu:
I realized this technically isn't guaranteed by the documentation, but it would be very surprising for users if `try_lock` randomly failed even if the Mutex was currently unlocked.

PR #210, opened 2020-01-22
Synchronize on compare_exchange failure in create_hashtable

tmiasko:


Amanieu:
Nice catch!

bors r+

tmiasko:
Somewhat as a side note: in C11/C++11 one could arguably use release as a success
ordering and acquire as a failure ordering, while Rust currently requires the
failure ordering to be equivalent or weaker. The C++20, lifts restrictions on
failure ordering completely http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0418r2.html.

Amanieu:
Could you open an issue on the Rust repo for lifting the restrictions on compare_exchange?

tmiasko:
https://github.com/rust-lang/rust/issues/68464

PR #217, opened 2020-02-18
Implement Support for creating the locking primitives in a const context on stable Rust

CryZe:
This adds a new constructor `const_new` to lock_api's mutexes, reentrant mutexes and RwLocks that allows creating them in a constant context on stable Rust by manually passing in the underlying raw mutex / rwlock.

Additionally, various helper functions are being added to parking_lot that allow creating mutexes, reentrant mutexes, fair mutexes and RwLocks in constant contexts on stable Rust.

### DIFF
--- a/lock_api/src/mutex.rs
+++ b/lock_api/src/mutex.rs
@@ -132,8 +132,9 @@ impl<R: RawMutex, T> Mutex<R, T> {
 }
 
 impl<R, T> Mutex<R, T> {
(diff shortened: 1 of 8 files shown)


CryZe:
This is a rough initial implementation. I'm not sure if the helpers in `parking_lot` are something that we would want as well. I can remove them if not.

thomcc:
Any chance this could get cut in a release? I could probably deprecate https://crates.io/crates/static_locks if it were.

PR #219, opened 2020-02-19
Remove the lazy_static dependency

CryZe:
I noticed there's a few things I missed in my previous PR.

1. lazy_static is not necessary anymore. We can just use the new const constructor functions instead.
2. The README still mentions lazy_static.


Amanieu:
bors retry

PR #223, opened 2020-04-09
Use llvm_asm! instead of asm!

Amanieu:
`asm!` will be deprecated soon in preparation for the new `asm!` macro, so switch to using `llvm_asm!` instead.

cc https://github.com/rust-lang/rfcs/pull/2843


Amanieu:
bors retry

PR #226, opened 2020-04-21
Replace Instant with dummy on wasm32

d3lm:


d3lm:
Ah wait! I suppose that `now()` should at least return a number, right?

d3lm:
Fixing it.

Cargo.toml
```diff
@@ -13,6 +13,7 @@ edition = "2018"
 [dependencies]
 parking_lot_core = { path = "core", version = "0.7.1" }
 lock_api = { path = "lock_api", version = "0.3.4" }
+cfg-if = "0.1"
```
Amanieu, line 16: This is not needed. `core/Cargo.toml` already has a dependency on `cfg-if`.

d3lm: Sorry, did not see that.

d3lm:
Alright. The PR is updated and it successfully builds.

core/src/parking_lot.rs
```diff
@@ -16,6 +16,39 @@ use core::{
...
+        struct DummyInstant(Duration);
+
+        impl DummyInstant {
+            pub fn now() -> DummyInstant {
+                DummyInstant::zero()
+            }
+
+            pub const fn zero() -> DummyInstant {
+                DummyInstant(Duration::from_secs(0))
+            }
```
Amanieu, line 33: This method is only used in one place, which means that:
- It doesn't need to be `pub`.
- You may as well just inline it into the `add` function.

d3lm: Yep, so I could reuse the logic there.

d3lm: Removing `pub`. Good point!

d3lm: I ll wait for Travis to finish to see if all goes green.

core/src/parking_lot.rs
```diff
@@ -16,6 +16,39 @@ use core::{
 use smallvec::SmallVec;
 use std::time::{Duration, Instant};
 
+cfg_if! {
+    if #[cfg(target_arch = "wasm32")] {
```
Amanieu, line 20: This should only be used on wasm32-unknown-unknown. Other wasm targets have a working `Instant` type which should be used.

d3lm: Yep! Changed it.

d3lm:
Yay, Travis goes green ✅

d3lm:
Updated 👍

core/src/parking_lot.rs
```diff
@@ -16,6 +16,39 @@ use core::{
 use smallvec::SmallVec;
 use std::time::{Duration, Instant};
 
+cfg_if! {
+    if #[cfg(target_arch = "wasm32-unknown-unknown")] {
```
Amanieu, line 20: That doesn't actually work, you need to check `target_os = "unknown"` and `target_vendor = "unknown"`.

d3lm: Oh ok, my bad.

core/src/parking_lot.rs
```diff
@@ -16,6 +16,39 @@ use core::{
...
+
+        impl Add<Duration> for DummyInstant {
+            type Output = DummyInstant;
+
+            fn add(self, _rhs: Duration) -> DummyInstant {
+                DummyInstant::zero()
+            }
+        }
+
+        // Use dummy implementation for `Instant` on `wasm32`
```
Amanieu, line 44: Add a small comment explaining that `Instant::now()` always panics on wasm32-unknown-unknown.

Amanieu:
Thanks!

d3lm:
Thank you 🙏 for being so active and responsive and also for providing this library. Really great work.

d3lm:
@Amanieu Will you release this today as well? Would be really cool.

PR #231, opened 2020-05-15
Uses `instant::Instant` in place of `std::time::Instant`.

azriel91:
Heya, I was trying out [`governor`](https://github.com/antifuchs/governor) (rate limiter) on WASM, which required `parking_lot` to use a wasm compatible `Instant` type, and I came up with this patch.

The underlying `instant::Instant` type is `std::time::Instant` when compiled without enabling the `"stdweb"` or `"wasm-bindgen"` features.

If you'd like to take it I'll be happy; I'm still trying out different rate-limiting implementations, so am not sure if I'll be using this (but if it's merged, then it's more likely that I will).

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -15,6 +15,7 @@ smallvec = "1.0"
 petgraph = { version = "0.5", optional = true }
 thread-id = { version = "3.2.0", optional = true }
 backtrace = { version = "0.3.2", optional = true }
+instant = "0.1"
 
 [target.'cfg(unix)'.dependencies]
 libc = "0.2.55"
--- a/core/src/parking_lot.rs
+++ b/core/src/parking_lot.rs
@@ -4,17 +4,18 @@
 // http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
-use cfg_if::cfg_if;
 use crate::thread_parker::{ThreadParker, ThreadParkerT, UnparkHandleT};
 use crate::util::UncheckedOptionExt;
 use crate::word_lock::WordLock;
+use cfg_if::cfg_if;
 use core::{
     cell::{Cell, UnsafeCell},
     ptr,
     sync::atomic::{AtomicPtr, AtomicUsize, Ordering},
 };
+use instant::Instant;
 use smallvec::SmallVec;
-use std::time::{Duration, Instant};
+use std::time::Duration;
 
 cfg_if! {
     if #[cfg(all(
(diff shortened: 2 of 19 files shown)


core/Cargo.toml
```diff
@@ -15,6 +15,7 @@ smallvec = "1.0"
 petgraph = { version = "0.5", optional = true }
 thread-id = { version = "3.2.0", optional = true }
 backtrace = { version = "0.3.2", optional = true }
+instant = "0.1"
```
Amanieu, line 18: I like the concept of `instant` but I am concerned that it is pulling in a lot of dependencies (through `time`) on non-WASM platforms. Looking at the code of that crate, it seems that `time` is only used to provide a separate `now` function. Could that be made optional so `parking_lot` doesn't have to depend on it?

core/src/parking_lot.rs
```diff
@@ -51,7 +52,7 @@ cfg_if! {
         // See https://github.com/rust-lang/rust/blob/master/src/libstd/sys/wasm/time.rs
         type InstantType = DummyInstant;
     } else {
-        // Otherwise use `std::time::Instant`
+        // Otherwise use `instant::Instant`
         type InstantType = Instant;
     }
 }
```
Amanieu, line 58: Is this hack still necessary now that we have support for Instant on WASM? I guess the default instant type is still used if neither the `stdweb` or `wasm-bindgen` features are enabled...

core/Cargo.toml
azriel91: hm, `time` is optional in `instant` by default (it needs the `"now"` feature enabled).
Is that sufficient? Also, should I wire through the `"now"` feature in `parking_lot`?

Amanieu: Ah sorry, I missed that. There's no need to wire through `"now"` since parking_lot doesn't use it.

core/src/parking_lot.rs
azriel91: I'm not sure, but shall check (give me an hour)

azriel91: It works without the hack :v:

GIF of a single threaded WASM countdown (counts down quickly, but only refreshes at 10 FPS):

wasm_countdown

azriel91:
Build failure on `x86_64-linux-android`:

```bash
error: linking with `cc` failed: exit code: 1
# ..
  = note: /usr/bin/ld: cannot find -llog
          collect2: error: ld returned 1 exit status
```

Not sure if it's related to this change -- `instant` doesn't introduce any system dependencies

Amanieu:
I think it's unrelated, maybe there is some issue with Android CI?

cc @XAMPPRocky

core/src/parking_lot.rs
Amanieu: After thinking about it for a bit, I think it's better to keep the old `DummyInstant` hack for wasm. The code using it relies on the fact that calling `Instant::now` is reasonably fast, which may not be the case on wasm.

azriel91: Gotcha, shall take take commit out

XAMPPRocky:
It looks like a regression in the nightly compiler to me. If it happens consistently I would report it to the main rust repo.

PR #233, opened 2020-05-17
Make RawReentrantMutex public

coolreader18:
This is useful for any use case that you'd be using a raw mutex for, where you also want reentrancy.

### DIFF
--- a/lock_api/src/remutex.rs
+++ b/lock_api/src/remutex.rs
@@ -47,14 +47,36 @@ pub unsafe trait GetThreadId {
     fn nonzero_thread_id(&self) -> NonZeroUsize;
 }
 
-struct RawReentrantMutex<R, G> {
+/// A raw mutex type that wraps another raw mutex to provide reentrancy.
+///
+/// Although this has the same methods as the [`RawMutex`] trait, it does
+/// not implement it, and should not be used in the same way, since this
+/// mutex can successfully acquire a lock multiple times in the same thread.
+/// Only use this when you know you want a raw mutex that can be locked
+/// reentrantly; you probably want [`ReentrantMutex`] instead.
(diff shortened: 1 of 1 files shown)


coolreader18:
I don't think the android build failure is related to my changes, right?

coolreader18:
Oh, hmm, this isn't sound, since when `RawMutex::lock()` succeeds `Mutex` assumes that it's completely exclusive. Maybe there could be another marker type on `RawMutex` to allow unifying `Mutex` and `ReentrantMutex` to the same generic type?

Amanieu:
You're right, you can't implement the `RawMutex` trait for a reentrant mutex since it would allow you to produce mutable `&mut` to the same data.

I don't think there is much of a use case for a raw reentrant mutex since you can always use `ReentrantMutex<()>` instead.

coolreader18:
Maybe `ReentrantMutex<()>` could expose an API that's similar to `RawMutex`? Idk, I probably have like the only use case for this that wouldn't need to use `unsafe` other than maybe `ReentrantMutex::force_unlock`; I'm implementing a `Lock` class for python, where all the other data structures are atomic/mutex-backed as well, but the lock is used for synchronization of multiple different operations that might modify multiple different data structures. The current implementation that I'm basing on this branch uses a trait with an associated `RawMutex` type, and I use that to implement reentrant and non-reentrant locks. I could probably just use a macro, though, or duplicate a bit of code.

coolreader18:
Maybe `RawReentrantMutex` could be made public with only the methods from the `RawMutex*` traits, but not actually implementing them?

Amanieu:
If you really want to expose it publicly then you should keep the methods as they are instead of using traits.

PR #235, opened 2020-05-18
Add RawMutex::is_locked()

### LINKED ISSUE #232: An is_locked API for RawMutex?
I'm using `parking_lot` and `lock_api` to implement the `_thread` Python module, and it works really well! The one thing that doesn't have a mirror is the `locked()` method of lock objects, which tells whether the lock is currently locked. Would it be feasible to add something like a `RawMutexStatus` trait that has an `is_locked(&self) -> bool` method, or would that not be able to be implemented for `parking_lot::RawMutex`, due to that not being a proper atomic operation or something?

coolreader18:
Resolves #232

### DIFF
--- a/lock_api/src/remutex.rs
+++ b/lock_api/src/remutex.rs
@@ -101,8 +101,7 @@ impl<R: RawMutex, G: GetThreadId> RawReentrantMutex<R, G> {
 
     #[inline]
     fn is_locked(&self) -> bool {
-        let id = self.get_thread_id.nonzero_thread_id().get();
-        self.owner.load(Ordering::Relaxed) == id || self.mutex.is_locked()
+        self.mutex.is_locked()
     }
 }
 
--- a/lock_api/src/remutex.rs
+++ b/lock_api/src/remutex.rs
@@ -98,6 +98,12 @@ impl<R: RawMutex, G: GetThreadId> RawReentrantMutex<R, G> {
             self.mutex.unlock();
         }
     }
+
+    #[inline]
+    fn is_locked(&self) -> bool {
+        let id = self.get_thread_id.nonzero_thread_id().get();
+        self.owner.load(Ordering::Relaxed) == id || self.mutex.is_locked()
+    }
 }
 
 impl<R: RawMutexFair, G: GetThreadId> RawReentrantMutex<R, G> {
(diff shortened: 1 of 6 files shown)


Amanieu:
For consistency, could you add a similar API to ReentrantMutex and RwLock?

coolreader18:
I'm not sure about the `RwLock` naming, but I think it's consistent with the difference between `RawRwLock::lock_{shared,exclusive}` and `RwLock::{read,write}`

lock_api/src/remutex.rs
```diff
@@ -98,6 +98,12 @@ impl<R: RawMutex, G: GetThreadId> RawReentrantMutex<R, G> {
             self.mutex.unlock();
         }
     }
+
+    #[inline]
+    fn is_locked(&self) -> bool {
+        let id = self.get_thread_id.nonzero_thread_id().get();
+        self.owner.load(Ordering::Relaxed) == id || self.mutex.is_locked()
```
Amanieu, line 105: This is unnecessary, you can just check `self.mutex.is_locked()`.

lock_api/src/rwlock.rs
```diff
@@ -56,6 +56,29 @@ pub unsafe trait RawRwLock {
...
     fn unlock_exclusive(&self);
+
+    /// Checks if this `RwLock` is currently locked in any way.
+    fn is_locked(&self) -> bool {
+        self.is_locked_shared() || self.is_locked_exclusive()
+    }
+
+    /// Checks if this `RwLock` currently has shared locks.
+    fn is_locked_shared(&self) -> bool {
+        let acquired_lock = self.try_lock_shared();
```
Amanieu, line 67: This won't work since you can acquire a read lock multiple times.

lock_api/src/rwlock.rs
```diff
@@ -56,6 +56,29 @@ pub unsafe trait RawRwLock {
 
     /// Releases an exclusive lock.
     fn unlock_exclusive(&self);
+
+    /// Checks if this `RwLock` is currently locked in any way.
+    fn is_locked(&self) -> bool {
+        self.is_locked_shared() || self.is_locked_exclusive()
```
Amanieu, line 62: I think we should only expose a single `is_locked` function here. I don't think there is much value in trying to differentiate between readers and writers.

Amanieu:
Missing an implementation of `is_locked` for `RawRwLock`. You should check if `WRITER_BIT | READERS_MASK` is non-zero.

Amanieu:
Thanks!

PR #240, opened 2020-05-27
Allow lock guards to be sent to other threads

### LINKED ISSUE #197: Sendable guards
> The deadlock detection code (which can be enabled through a feature) will break if a mutex can be unlocked in a different thread than the one that locked it. It was decided that disabling `MutexGuard: Send` if that feature is enabled would be unacceptable since it might break code that relies on `MutexGuard: Send`.

_Originally posted by @Amanieu in https://github.com/Amanieu/parking_lot/issues/139#issuecomment-490236259_

While I understand the reasoning in the quote above, it's not ideal. This a bit weird for me to say, considering I proposed and implemented the PR that introduced the deadlock detection :rofl: 

Parkinglot is the **only** crate that _can/could_ do Sendable guard AFAIK. So maybe should try to  preserve it somehow?

Sendable guards may come in handy, as it just did for me, and you'd have to use an unsafe impl to work around it and hope that nobody enables deadlock detection.

Some ideas, each with its own {up/down}sides
1 - Reintroduce Sendable guards which is implicitly a `!deadlock` feature.
2 - Create a Sendable{Mutex,RwLock} or something like that that don't participate in the deadlock detection.
3 - Same as above, but in another crate.
4 - ..

Amanieu:
But only if deadlock detection is not enabled.

Fixes #197

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,8 @@ in the Rust standard library:
 18. Optional support for [serde](https://docs.serde.rs/serde/).  Enable via the
     feature `serde`.  **NOTE!** this support is for `Mutex`, `ReentrantMutex`,
     and `RwLock` only; `Condvar` and `Once` are not currently supported.
+19. Lock guards can be sent to other threads when the `send_guard` geature is
+    enabled.
 
 ## The parking lot
 
@@ -117,6 +119,12 @@ parking_lot = { version = "0.10", features = ["nightly"] }
 
 The experimental deadlock detector can be enabled with the
 `deadlock_detection` Cargo feature.
+
+To allow sending `MutexGuard`s and `RwLock*Guard`s to other threads, enable the
+`send_guard` option.
+
+Note that the `deadlock_detection` and `send_guard` features are incompatible
+and cannot be used together.
 
 The core parking lot API is provided by the `parking_lot_core` crate. It is
 separate from the synchronization primitives in the `parking_lot` crate so that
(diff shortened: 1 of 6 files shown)


Amanieu:
cc @arthurprs

bjorn3:
Cargo expects features to be additive. This means that enabling features must **never** break code. It must only allow more code. By making the deadlock detection feature disable the `Send` impl for `MutexGuard` you are breaking code.

arthurprs:
While that's the expectation, there are lots and lots of examples of mutually exclusive features out there. But that's just my opinion.

bjorn3:
In those cases it is often choosing between two or more backends for a crate. This means that 1. the target executable is likely the one enabling the feature flag and 2. you can't accidentally depend on a feature flag being disabled, as you have to explicitly enable one feature flag to break another one. You can't just have a crate compiling with no feature flags enabled at all and then break when you enable one feature flag.

bjorn3:
A better alternative than simply breaking code when enabling the `deadlock_detection` feature flag would be to add a new `sendable_guard` feature flag to enable this and add a `compile_error!` when both feature flags are enabled. While this does still violate the additive assumption, this does makes it a lot harder to accidentally depend on the `Send`ness of lock guards.

Amanieu:
@bjorn3 That sounds like a great idea!

README.md
```diff
@@ -72,6 +72,8 @@ in the Rust standard library:
 18. Optional support for [serde](https://docs.serde.rs/serde/).  Enable via the
     feature `serde`.  **NOTE!** this support is for `Mutex`, `ReentrantMutex`,
     and `RwLock` only; `Condvar` and `Once` are not currently supported.
+19. Lock guards can be sent to other threads when the `send_guard` geature is
```
bjorn3, line 75:
```suggestion
19. Lock guards can be sent to other threads when the `send_guard` feature is
```

PR #241, opened 2020-06-06
Updating cloudabi to 0.1

mettke:
Hey folks, 

cloudabi was updated to 0.1. The commit https://github.com/NuxiNL/cloudabi/commit/15f18a92925605b8edd278ed049f2520f1fd56f1 is mainly about updating their bitflags dependency thus this crate should be safe to update too.


Amanieu:
Thanks!

PR #247, opened 2020-07-04
lock_api: Allow unsafely accessing data without a guard

### LINKED ISSUE #246: lock_api: Unsafely access data within a `Mutex` without `MutexGuard`
I have a use-case where I need to manually manage locking/unlocking a `Mutex`, and can't store the `MutexGuard` between locking and accessing the locked data. It's possible to perform locking and unlocking of the `RawMutex` manually from outside this crate using the `Mutex::raw()` method, as well as `mem::forget(mutex.lock())` and `mutex.force_unlock()`, but there appears to be no API to then unsafely access the raw data guarded by a `&Mutex`.

It would be nice to expose a method to extract the raw pointer to the mutex-guarded data within the `UnsafeCell`, perhaps with something like the following:

```rust
/// Returns a raw pointer to the underlying data.
///
/// # Safety
///
/// The returned pointer may only be dereferenced if the current thread logically
/// owns a `MutexGuard`, but that guard has been discarded using `mem::forget`.
pub fn data_ptr(&self) -> *mut T {
  self.data.get()
}
```

mystor:
Fixes #246

### DIFF
--- a/lock_api/src/mutex.rs
+++ b/lock_api/src/mutex.rs
@@ -279,9 +279,9 @@ impl<R: RawMutex, T: ?Sized> Mutex<R, T> {
     ///
     /// # Safety
     ///
-    /// The returned pointer must only be dereferenced if the current thread
-    /// logically owns a `MutexGuard` but that guard has been discarded using
-    /// `mem::forget`.
+    /// You must ensure that there are no data races when dereferencing the
+    /// returned pointer, for example if the current thread logically owns
+    /// a `MutexGuard` but that guard has been discarded using `mem::forget`.
     #[inline]
     pub fn data_ptr(&self) -> *mut T {
         self.data.get()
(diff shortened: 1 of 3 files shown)


lock_api/src/mutex.rs
```diff
@@ -270,6 +270,22 @@ impl<R: RawMutex, T: ?Sized> Mutex<R, T> {
...
+    ///
+    /// This is useful when combined with `mem::forget` to hold a lock without
+    /// the need to maintain a `MutexGuard` object alive, for example when
+    /// dealing with FFI.
+    ///
+    /// # Safety
+    ///
+    /// The returned pointer must only be dereferenced if the current thread
+    /// logically owns a `MutexGuard` but that guard has been discarded using
+    /// `mem::forget`.
```
Amanieu, line 284:
```suggestion
    /// You must ensure that there are no data races when dereferencing the
    /// returned pointer, for example if the current thread logically owns
    /// a `MutexGuard` but that guard has been discarded using `mem::forget`.
```

Amanieu:
Thanks!

PR #252, opened 2020-10-03
Fix for two Miri issues with unit tests

m-ou-se:
Fixes https://github.com/Amanieu/parking_lot/issues/251

With this change, all parking_lot tests pass with Miri with https://github.com/rust-lang/miri/pull/1568 applied, except for the `webkit_queue_test`s which are too slow to complete in reasonable time with Miri. (Lowering the number of messages per producer to 100 or 10 makes them pass too.)

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -42,7 +42,6 @@ jobs:
           - wasm32-unknown-unknown
           - x86_64-fortanix-unknown-sgx
           #- x86_64-unknown-redox
-          - x86_64-unknown-cloudabi
           #- x86_64-linux-android
     steps:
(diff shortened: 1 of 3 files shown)


m-ou-se:
The failing test is unrelated:

```
error: component 'rust-std' for target 'x86_64-unknown-cloudabi' is unavailable for download for channel nightly
```

`rust-std` for CloudABI is no longer distributed by rustup. (https://github.com/rust-lang/rust/pull/75568)

Amanieu:
Just remove cloudabi from .travis.yml

Amanieu:
Oh wait there's no bors here.

PR #254, opened 2020-10-19
Update lock_api and instant to most recent versions

j-vanderstoep:
lock_api is updated from 0.4.0 to 0.4.1.
instant is upsated from 0.1.4 to 0.1.7.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,8 +12,8 @@ edition = "2018"
 
 [dependencies]
 parking_lot_core = { path = "core", version = "0.8.0" }
-lock_api = { path = "lock_api", version = "0.4.0" }
-instant = "0.1.4"
+lock_api = { path = "lock_api", version = "0.4.1" }
+instant = "0.1.7"
 
 [dev-dependencies]
 rand = "0.7.3"


j-vanderstoep:
Thanks for maintaining this crate. Some additional context for this change:

We're considering parking_lot for use in the Android open source project [1]. For Android, it's important that we limit the number of crate versions that are used because that reduces both memory and disk usage, both of critical importance on resource-constrained Android devices. Generally we try to achieve that by only depending on the latest version of each crate that we directly or indirectly use. This change moves us to the most recent version of lock_api and instant crates.

[1] https://android-review.googlesource.com/c/platform/external/rust/crates/parking_lot/+/1460364

CryZe:
This should not be necessary, Rust will never select 0.1.4 and 0.1.7 as incompatible crates were it would compile in both versions.

j-vanderstoep:
> This should not be necessary, cargo will never select both 0.1.4 and 0.1.7 as incompatible crates were it would compile in both versions.

I'm not sure I understand what you mean. If project A depends in 0.1.4, and project B depends on version 0.1.7, cargo will pull in 0.1.4 for project A and 0.1.7 for project B. Which would result in two versions of the same dylib on my device. If instead they both depend on 0.1.7, then I can have both projects depend on the same compiled output.

CryZe:
That's not the case, if both 0.1.4 and 0.1.7 are depended on cargo will only pull in 0.1.7, unless they both specify them as "=0.1.4" and "=0.1.7" respectively.

j-vanderstoep:
Isn't that only true if 0.1.4 and 0.1.7 are depended on by the same project (e.g. a single "cargo build" command)? I'm not just concerned about a single project and its dependencies. I want separate/independent projects to rely on the same version of the crate if possible. Or does cargo always pull in the most recent minor version?

bjorn3:
The `Cargo.lock` file is used to ensure that the used version of dependencies doesn't change between builds. You can use `cargo update` to update all dependencies or `cargo update -p lock_api` to only update `lock_api`.

faern:
@j-vanderstoep When you do `cargo update` it will automatically upgrade all crates to their latest semver compatible versions, depending on how the version in specified in `Cargo.toml`. Here in `parking_lot` the version is specified as `version = "0.4.0"` (one example). This is equivalent to writing `version = "^0.4.0"`. It's called a caret requirement. This allows cargo to pull in any version in this range: `>=0.4.0, <0.5.0`. Please see the relevant cargo documentation: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

So yes, the change the PR makes should not be needed. You should just be able to do `cargo update` to get the latest versions. As pointed out above you can run it with `-p <crate>` to only update some crates.

j-vanderstoep:
Thanks everyone for the explanation. I confirmed locally exactly what you're describing - Even if 0.1.4 is specified, if 0.1.7 exists, Cargo will use that instead.

PR #256, opened 2020-10-24
Add winapi srwlock to benchmarks

nico-abram:
Since I saw you had pthreads in the benchmarks I figured it would make sense to add windows slim rwlocks to it too

It seems std uses srwlocks internally, but it might still make sense to add this since

 1) std could change in the future
 2) It has a fallback for windows xp which looks like it would possibly have some runtime overhead (Even if perfectly predicted the branch could still degrade performance since it would take a spot in the branch predictor right?)
 3) I see you also have pthread implementations in the benchmarks despite std using pthreads internally

With regards to point (2), I see non trivial differences between srwlock and std in an uncontended case:
imagen

### DIFF
--- a/benchmark/Cargo.toml
+++ b/benchmark/Cargo.toml
@@ -20,3 +20,6 @@ path = "src/rwlock.rs"
 [features]
 nightly = ["parking_lot/nightly"]
 deadlock_detection = ["parking_lot/deadlock_detection"]
+
+[target.'cfg(windows)'.dependencies]
+winapi = { version = "0.3", features = ["synchapi"] }
--- a/benchmark/src/mutex.rs
+++ b/benchmark/src/mutex.rs
@@ -8,7 +8,7 @@
 mod args;
 use crate::args::ArgRange;
 
-#[cfg(unix)]
+#[cfg(any(windows, unix))]
 use std::cell::UnsafeCell;
 use std::{
     sync::{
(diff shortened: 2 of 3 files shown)


benchmark/Cargo.toml
```diff
@@ -20,3 +20,6 @@ path = "src/rwlock.rs"
 [features]
 nightly = ["parking_lot/nightly"]
 deadlock_detection = ["parking_lot/deadlock_detection"]
+
+[target.'cfg(windows)'.dependencies]
+winapi = { version = "0.3", features = ["synchapi"] }
```
bjorn3, line 25: nit: missing trailing newline

nico-abram: Thanks! Solved

nico-abram:
Force pushed trailing newline and missing type declaration in non windows because of the if cfg!

benchmark/src/mutex.rs
```diff
@@ -57,6 +57,55 @@ impl<T> Mutex<T> for parking_lot::Mutex<T> {
...
+        }
+    }
+    fn name() -> &'static str {
+        "winapi_srwlock"
+    }
+}
+#[cfg(windows)]
+impl<T> Drop for SrwLock<T> {
+    fn drop(&mut self) {
+        // TODO: Do we need to do anything here? CloseHandle?
```
Amanieu, line 102: Nothing is needed here. You can remove the TODO.

nico-abram: Force pushed this change

PR #262, opened 2020-11-17
Fix bounds on Send and Sync impls for lock guards

### LINKED ISSUE #258: Mapped*Guard objects should only be Send when the contained type is
When using Sendable guard objects, such as with the `send_guard` feature or the locks in the [spin](https://crates.io/crates/spin) crate, it is possible to get non-Sendable types across threads using the `Mapped*Guard` (`MappedMutexGuard`, `MappedRwLockReadGuard`, `MappedRwLockWriteGuard`) objects.

Currently, they're marked as `Send` as long as the raw lock is `Send` with no bound on `T`:
https://github.com/Amanieu/parking_lot/blob/5ac2971da40006f678fa78117201b7521e9df1fd/lock_api/src/mutex.rs#L604-L607

This should probably have a `T: Send` bound, just like the regular [`MutexGuard` does](https://docs.rs/lock_api/0.4.0/lock_api/struct.MutexGuard.html#impl-Send). (Issue found by @sslab-gatech's Rust group)

----

Here's a proof-of-concept compiled with 

```toml
parking_lot = { version = "=0.11.0", features = ['send_guard']}
```
using this issue to cause a data-race with a `Cell` object

```rust
#![forbid(unsafe_code)]
use std::cell::Cell;

use parking_lot::{Mutex, MutexGuard};
use crossbeam_utils::thread;

#[derive(Debug, Clone, Copy)]
enum RefOrInt<'a> {
    Ref(&'a u64),
    Int(u64),
}
static SOME_INT: u64 = 123;

fn main() {
    let cell = Cell::new(RefOrInt::Ref(&SOME_INT));
    let mutex = Mutex::new(&cell);

    thread::scope(|s| {
        let guard = mutex.lock();
        // MappedMutexGuard that just returns the whole object as a "component".
        let mapped_guard = MutexGuard::map(guard, |x| x);

        let child = s.spawn(move |_| {
            let smuggled_cell = *mapped_guard;

            println!("Thread - {:p} - {:?}", smuggled_cell, smuggled_cell);
            loop {
                // Repeatedly write Ref(&addr) and Int(0xdeadbeef) into the cell.
                smuggled_cell.set(RefOrInt::Ref(&SOME_INT));
                smuggled_cell.set(RefOrInt::Int(0xdeadbeef));
            }
        });

        println!("Main - {:p} - {:?}", &cell, cell);
        loop {
            if let RefOrInt::Ref(addr) = cell.get() {
                // Hope that between the time we pattern match the object as a
                // `Ref`, it gets written to by the other thread.
                if addr as *const u64 == &SOME_INT as *const u64 {
                    continue;
                }

                // Due to the data race, obtaining Ref(0xdeadbeef) is possible
                println!("Pointer is now: {:p}", addr);
                println!("Dereferencing addr will now segfault: {}", *addr);
            }
        }
    });
```

This outputs:
```
Main - 0x7ffe6021e8a8 - Cell { value: Ref(123) }
Thread - 0x7ffe6021e8a8 - Cell { value: Ref(123) }
Pointer is now: 0xdeadbeef

Return Code: -11 (SIGSEGV)
```

### LINKED ISSUE #259: RwLockReadGuard and RwLockWriteGuard should only be Sync when RawRwLock is
Currently `RwLockReadGuard`s are marked as `Sync` so long as the contained type is, without having the same bound on the `RawRwLock` type:

https://github.com/Amanieu/parking_lot/blob/ecfe20bf5b7f8b06c7a39afd53cee88a99ce07db/lock_api/src/rwlock.rs#L878

This means that even if an implementer marks their guard as `GuardNoSend`, it's actually possible to send guard objects across threads using references to them (references are marked as `Send` so long as the underlying type is `Sync`).

Much like `MappedMutexGuard` this should probably only be `Sync` when the raw lock is.

An example of sending `parking_lot`'s non sendable guard objects across threads:
```rust
use parking_lot::RwLock;
use crossbeam_utils::thread;

fn is_send<T: Send>(_: &T) {}

fn main() {
    let lock = RwLock::new(42);
    let read_guard = lock.read();
    // RwLockReadGuard itself is not Send-able, uncommenting
    // the line below causes a compilation failure.
    //is_send(&read_guard);
    
    thread::scope(|s| {
        let guard_ref = &read_guard;
        
        s.spawn(move |_| {
            // But they can be sent as a reference.
            let sent_lock_guard = guard_ref;
            println!("In thread: {:p}", sent_lock_guard);
        });
        println!("In main:   {:p}", &read_guard);
    });
}
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=625dd71fdcd34658f0c2d2752a86ca0e

(Issue found by @sslab-gatech's Rust group)

Amanieu:
Fixes #258
Fixes #259

### DIFF
--- a/lock_api/src/lib.rs
+++ b/lock_api/src/lib.rs
@@ -99,6 +99,8 @@ pub struct GuardSend(());
 /// Marker type which indicates that the Guard type for a lock is not `Send`.
 pub struct GuardNoSend(*mut ());
 
+unsafe impl Sync for GuardNoSend {}
+
 mod mutex;
 pub use crate::mutex::*;
 
--- a/lock_api/src/mutex.rs
+++ b/lock_api/src/mutex.rs
@@ -601,7 +601,7 @@ unsafe impl<'a, R: RawMutex + Sync + 'a, T: ?Sized + Sync + 'a> Sync
     for MappedMutexGuard<'a, R, T>
 {
 }
-unsafe impl<'a, R: RawMutex + 'a, T: ?Sized + 'a> Send for MappedMutexGuard<'a, R, T> where
+unsafe impl<'a, R: RawMutex + 'a, T: ?Sized + Send + 'a> Send for MappedMutexGuard<'a, R, T> where
     R::GuardMarker: Send
 {
 }
--- a/lock_api/src/rwlock.rs
+++ b/lock_api/src/rwlock.rs
@@ -875,8 +875,6 @@ pub struct RwLockReadGuard<'a, R: RawRwLock, T: ?Sized> {
     marker: PhantomData<(&'a T, R::GuardMarker)>,
 }
 
-unsafe impl<'a, R: RawRwLock + 'a, T: ?Sized + Sync + 'a> Sync for RwLockReadGuard<'a, R, T> {}
-
 impl<'a, R: RawRwLock + 'a, T: ?Sized + 'a> RwLockReadGuard<'a, R, T> {
     /// Returns a reference to the original reader-writer lock object.
     pub fn rwlock(s: &Self) -> &'a RwLock<R, T> {
@@ -1050,8 +1048,6 @@ pub struct RwLockWriteGuard<'a, R: RawRwLock, T: ?Sized> {
     rwlock: &'a RwLock<R, T>,
     marker: PhantomData<(&'a mut T, R::GuardMarker)>,
 }
-
-unsafe impl<'a, R: RawRwLock + 'a, T: ?Sized + Sync + 'a> Sync for RwLockWriteGuard<'a, R, T> {}
 
 impl<'a, R: RawRwLock + 'a, T: ?Sized + 'a> RwLockWriteGuard<'a, R, T> {
     /// Returns a reference to the original reader-writer lock object.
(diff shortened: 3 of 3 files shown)


ammaraskar:
Thank you! ❤️

RalfJung:
I have looked at this on our formal soundness proof of (idealized) Mutex and RwLock implementations. I can confirm that `T: Send` implies `MutexGuard<T>: Send`.

However, for `RwLock`, things are more tricky. The way we have set up the proofs, we need *both* `T: Send` and `T: Sync` to show that either of the guards is `Send`. However, I think that is just an artifact of out proof -- I agree that these ought to be sound.

jhjourdan:
I had a more precise look on the formal proofs than @RalfJung did, and it turns out that indeed, it is safe to:
- make the read guard both `Send` and `Sync` as soon as `T : Sync`,
- make the write guard `Send` as soon as `T : Send`,
- make the write guard `Sync` as soon as `T : Sync`.

RalfJung:
IOW, the read guard behaves like a shared reference and the write guard behaves like a mutable reference. Exactly what one would expect. Thanks @jhjourdan!

ammaraskar:
Thanks for double-checking and formally proving this Ralf and Jacques. Good to see that this matches the intuition from https://github.com/Amanieu/parking_lot/issues/258#issuecomment-723671395

PR #265, opened 2020-12-04
Remove instant dummy and update to latest wasm32 simd intrinsics

d3lm:
This PR removes the `Instant` dummy implementation, which is no longer needed as we are using `instant` now instead of `std`. 

I have also added a polyfill for `Instant::checked_duration_since` because it's not yet implemented for the wasm32. Once it gets implemented we can remove it again.

Lastly, I have updated the `thread_parker` to use the latest wasm32 SIMD intrinsics. For reference, check [this](https://github.com/rust-lang/stdarch/commit/56ee49774172a40883c8c63508d1ec93b3d3c7ea#diff-4cad5669e37a1690249bffa95f51506b35cb27dad65188d63e020784c66fa4c0).

### DIFF
--- a/core/src/parking_lot.rs
+++ b/core/src/parking_lot.rs
@@ -7,7 +7,6 @@
 use crate::thread_parker::{ThreadParker, ThreadParkerT, UnparkHandleT};
 use crate::util::UncheckedOptionExt;
 use crate::word_lock::WordLock;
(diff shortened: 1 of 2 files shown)


d3lm:
I also ran this locally against my project and all 350 tests pass. Looking good.

PR #266, opened 2020-12-04
core: Use a weaker dependency on smallvec.

emilio:
Smallvec 1.4 regresses performance in Firefox, so while we fix it it'd
be nice to be able to not use the new version. parking_lot is not using
try_reserve so it can live with 1.0.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -11,7 +11,7 @@ edition = "2018"
 
 [dependencies]
 cfg-if = "1.0.0"
-smallvec = "1.4.0"
+smallvec = "1.0"
 petgraph = { version = "0.5.1", optional = true }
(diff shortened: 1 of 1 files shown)


emilio:
@Amanieu if you could do a dot-release with this change it'd be amazing. I checked 1.0.0 builds parking-lot just fine.

Amanieu:
Can't you fix the performance regression in smallvec itself?

emilio:
Yes, and I intend to do that. But we're about to cut a release and this seemed harmless in your side and avoids me having to patch smallvec to revert the offending patch for now.

Amanieu:
Published parking_lot_core 0.8.1.

emilio:
Thanks so much!

PR #270, opened 2020-12-19
OpenBSD returns EINVAL in the same way as DragonFlyBSD.

ltratt:
Without this, `cargo test` fails on OpenBSD -current. I suspect all the BSDs have the same behaviour in this regard, but I don't have the ability to test FreeBSD or NetBSD I'm afraid.


Amanieu:
I think we should just get rid of the `if` and always use the first variant. It's not really essential anyways.

ltratt:
I agree -- the two new commits do that. I'm happy to squash them down to one if you'd prefer.

Amanieu:
Thanks!

PR #276, opened 2021-01-27
Update smallvec to 1.6

### LINKED ISSUE #274: cargo audit alert for smallvec 1.6.0
Hi, would you please check cargo audit. I receive a message regarding one of parking_lot_core deps. and it would be nice to put the cargo audit in your build pipeline.
```bash
Crate:         smallvec
Version:       1.6.0
Title:         Buffer overflow in SmallVec::insert_many
Date:          2021-01-08
ID:            RUSTSEC-2021-0003
URL:           https://rustsec.org/advisories/RUSTSEC-2021-0003
Solution:      Upgrade to >=0.6.14, <1.0.0 OR >=1.6.1
```

FintanH:
Fixes #274 

There was a vulnerability found in smallvec as described in:
* https://github.com/servo/rust-smallvec/issues/252
* https://github.com/RustSec/advisory-db/pull/552

This patch updates the package version to 1.6 which is deemed safe to
use.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -11,7 +11,7 @@ edition = "2018"
 
 [dependencies]
 cfg-if = "1.0.0"
-smallvec = "1.0"
+smallvec = "1.6.1"
 petgraph = { version = "0.5.1", optional = true }
 thread-id = { version = "3.3.0", optional = true }
 backtrace = { version = "0.3.49", optional = true }


Amanieu:
You might want to specifically use 1.6.1 since I believe 1.6.0 doesn't contain the fix,

FintanH:
Ah, you're right! :facepalm:

rohitjoshi:
@Amanieu thanks

Issue #278, opened 2021-02-05
Possible bug - UB


Kixunil:
Based on [recent reddit post](https://www.reddit.com/r/rust/comments/lc9122/help_interpreting_a_segfault_in_tokio/) there may be a bug in `CondVar::wait_until_internal()`. Sadly the OP didn't provide more information and I didn't find anything wrong in the code. I didn't check too deeply so it'd be nice if someone else tried to check too if I missed something.

Amanieu:
`ThreadData::new` calls `grow_hashtable` which is a pretty large chunk of unsafe code. That's the only thing I can think of that could potentially segfault in that function.

Kixunil:
Exactly that was my thinking. I was suspecting hashmap being empty when created first but that's not actually the case since it's always created with three elements. Maybe testing that part of code under fuzzer and miri would help.

jefshe:
OP here! I've reproduced this bug in debug mode with the following stacktrace:
```
#0  0x00007f1eca5f9a9b in __memmove_avx_unaligned_erms () at ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:309
#1  0x000055a0b5cc4ab9 in core::intrinsics::copy_nonoverlapping (src=0x7f1d6affb508, dst=0x7f1ec5cf4120 <inflate_fast+64>, count=1) at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/core/src/intrinsics.rs:1860
#2  0x000055a0b5cc6c73 in core::ptr::swap_nonoverlapping_one (x=0x7f1ec5cf4120 <inflate_fast+64>, y=0x7f1d6affb508) at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/core/src/ptr/mod.rs:455
#3  0x000055a0b5cc4e24 in core::mem::swap (x=0x7f1ec5cf4120 <inflate_fast+64>, y=0x7f1d6affb508) at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/core/src/mem/mod.rs:688
#4  0x000055a0b5cc4f9c in core::mem::replace (dest=0x7f1ec5cf4120 <inflate_fast+64>, src=0x0) at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/core/src/mem/mod.rs:815
[...]

PR #279, opened 2021-02-10
Track more safety invariants in the type system

Kixunil:
## Meta - about this PR

This is a byproduct of attempting to figure out #278 by "asking the compiler if it sees a problem".
The idea seemed to be possibly useful, so this should be viewed almost like a draft/suggestion.

## Description

This change creates several wrappers around `unsafe` operations to
better track the safety invariants. While not perfect, which should
reduce the number of footguns in the code.

It focuses on locking - using primitives like mutex guards known from
`std`. The locked state is tracked in those guards. This crate needs
some special features however. It needs to lock all locks in a slice and
unlock them in the same order. Similarly it needs to lock a queue inside
a `Bucket` while keeping reference to the bucket.

Additional specialized wrappers maintain safety of these advanced
operations similarly to guards. They are implemented in their own
modules to avoid access from other parts of code.

This change comes with two more side effects:

* `Cell` was removed from the bucket as now locking provides required
  safety.
* Bucket pair uses `Option` to represent same buckets. This adds a
  slight performance penalty but it's in slow path and may be optimized
  out by the compiler. It's even possible that absence of `Cell` will
  enable even more optimizations.

### DIFF
--- a/core/src/parking_lot.rs
+++ b/core/src/parking_lot.rs
@@ -6,7 +6,6 @@
 // copied, modified, or distributed except according to those terms.
 use crate::thread_parker::{ThreadParker, ThreadParkerT, UnparkHandleT};
 use crate::util::UncheckedOptionExt;
-use crate::word_lock::WordLock;
 use core::{
     cell::{Cell, UnsafeCell},
     ptr,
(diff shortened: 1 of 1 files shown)


Kixunil:
Ah, this has lower MSRV. I'm too exhausted to fix it now. Will come back later.

Some more ideas to consider:
* Implement queue operations as methods on `Queue`
* Add `BucketPair` wrapper type holding the pair, providing a nice getter and ensuring correct drop order. (unrelated to #278 but nice to have)

PR #280, opened 2021-02-18
Add [Raw]ReentrantMutex::is_owned

coolreader18:
I'm not sure about the naming; I based it off of Python's RLock._is_owned, but Rust obviously has a pretty different meaning of "owned". Maybe a more explicit/verbose name, like `owned_by_thread`?


Amanieu:
Since this is a rather uncommon operation, I'd go for the full `is_owned_by_current_thread`.

coolreader18:
Would you be able to cut a new version with this change?

Amanieu:
Done.

PR #282, opened 2021-04-27
avoid deprecated spin_loop_hint

RalfJung:
I noticed some warnings when experimenting with using patched parking_lot in rustc... though somehow the warnings do not show for regular rustc builds?


RalfJung:
Oh, I guess this would raise the MSRV to 1.49 which you probably don't want to do yet.

PR #287, opened 2021-06-07
Optionally implement `bytemuck::Zeroable`, for safe zero-allocation

### LINKED ISSUE #286: Implement `bytemuck` traits where appropriate
A nice optimisation provided by `bytemuck` is the ability to do zeroed heap allocations using `calloc` safely. Since the default `parking_lot` `Mutex` and `RwLock` have an all-zeroes default state, then this means for any safely zeroable type `T`, `RwLock<T>` and `Mutex<T>` are also safely zeroable - and, by extension, `FairMutex` too. `ReentrantMutex` is not, however. Implementing these traits - even if behind a feature gate - would make some use-cases using collections of mutexes faster.

Vurich:
Closes #286.

We only use `derive(Zeroable)` in one place, as the atomic integer types don't implement `bytemuck::Zeroable` at time of writing. This appears to be an oversight in `bytemuck`.

The `Zeroable` trait only specifies that the type can be _created_ with an all-zeroes bit pattern, and so this cannot be abused to overwrite the locks with all-zeroes (although doing so would require `&mut` access anyway).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ edition = "2018"
 [dependencies]
 parking_lot_core = { path = "core", version = "0.8.0" }
 lock_api = { path = "lock_api", version = "0.4.0" }
+bytemuck = { version = "1.5.1", optional = true, features = ["derive"] }
 instant = "0.1.4"
 
 [dev-dependencies]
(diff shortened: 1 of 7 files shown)


Amanieu:
Do you think you could send a PR to bytemuck to support atomic types? That way we can just use `#[derive(Zeroable)]` everywhere.

Amanieu:
Can you document the feature in README. It's a bit confusing that we have both `bytemuck` and `bytemuck_traits` features, but that's an unfortunate cargo limitation.

bjorn3:
Having the feature called "bytemuck" works on nightly with `-Znamespaced-features` through `bytemuck = ["dep:bytemuck", "lock-api/bytemuck"]` I believe.

nolanderc:
`Zeroable` has been implemented for `Atomic*` types in https://github.com/Lokathor/bytemuck/pull/157

PR #290, opened 2021-06-21
Allow owned data in MappedMutexGuard

### LINKED ISSUE #289: Consider allowing owned data in `MappedMutexGuard`
There is no reason that `MappedMutexGuard` needs to contain a `*mut T`, it can just contain a `T`. I believe that you may have to change the bound of `map` like so:

```rust
pub fn map<U: ?Sized + 'a, F>(s: Self, f: F) -> MappedMutexGuard<'a, R, U>
where
    F: FnOnce(&'a mut T) -> U,
{
    let raw = &s.mutex.raw;
    let data = f(unsafe { &mut *s.mutex.data.get() });
    mem::forget(s);
    MappedMutexGuard {
        raw,
        data,
        marker: PhantomData,
    }
}
```

And you should also bound `MappedMutexGuard`'s `T` parameter by `'a`. This is strictly more general than the existing method, but will break anyone who explicitly specifies the `T` parameter of `MappedMutexGuard`. To get around this, you could rename `MappedMutexGuard` to `OwnedMappedMutexGuard` and make the original name a type alias like so:

```rust
pub type MappedMutexGuard<'a, R, T> = OwnedMappedMutexGuard<'a, R, &'a mut T>;
```

Vurich:
Closes #289


lock_api/src/mutex.rs
```diff
@@ -434,21 +434,23 @@ impl<'a, R: RawMutex + 'a, T: ?Sized + 'a> MutexGuard<'a, R, T> {
...
-    pub fn map<U: ?Sized, F>(s: Self, f: F) -> MappedMutexGuard<'a, R, U>
+    pub fn map<U: 'a, F>(s: Self, f: F) -> OwnedMappedMutexGuard<'a, R, U>
     where
-        F: FnOnce(&mut T) -> &mut U,
+        F: FnOnce(&'a mut T) -> U,
     {
         let raw = &s.mutex.raw;
-        let data = f(unsafe { &mut *s.mutex.data.get() });
+        let data = unsafe { &mut *s.mutex.data.get() };
         mem::forget(s);
```
bjorn3, line 443: This will invalidate the `data` reference.

lock_api/src/mutex.rs
```diff
@@ -458,17 +460,21 @@ impl<'a, R: RawMutex + 'a, T: ?Sized + 'a> MutexGuard<'a, R, T> {
...
-        F: FnOnce(&mut T) -> Option<&mut U>,
+        F: FnOnce(&'a mut T) -> Option<U>,
     {
+        let s = mem::ManuallyDrop::new(s);
+
         let raw = &s.mutex.raw;
-        let data = match f(unsafe { &mut *s.mutex.data.get() }) {
+        let data = unsafe { &mut *s.mutex.data.get() };
+
+        let data = match f(data) {
```
bjorn3, line 472: This will leak the mutex guard if `f(data)` panics.

lock_api/src/mutex.rs
```diff
@@ -591,79 +597,93 @@ unsafe impl<'a, R: RawMutex + 'a, T: ?Sized + 'a> StableAddress for MutexGuard<'
...
+        let mut s = mem::ManuallyDrop::new(s);
+
         let raw = s.raw;
-        let data = f(unsafe { &mut *s.data });
-        mem::forget(s);
-        MappedMutexGuard {
+
+        let data = unsafe { ptr::read(&mut s.data) };
+
+        let data = f(data);
```
bjorn3, line 643: Same

lock_api/src/mutex.rs
```diff
@@ -591,79 +597,93 @@ unsafe impl<'a, R: RawMutex + 'a, T: ?Sized + 'a> StableAddress for MutexGuard<'
...
-        };
-        mem::forget(s);
-        Ok(MappedMutexGuard {
-            raw,
-            data,
-            marker: PhantomData,
-        })
+        let data = unsafe { ptr::read(&mut s.data) };
+
+        f(data)
```
bjorn3, line 672: Same

Vurich:
@bjorn3 Thank you for your review. I've also noticed that I accidentally based this PR on my other PR implementing the `bytemuck` traits instead of on master, so I'll fix that too. Would you say that you're in favour of the general idea of the PR, even if there are issues with the implementation? I believe that it's a minor breaking change since it will affect anyone explicitly specifying the type and lifetime parameters of `MappedMutexGuard::map`/`try_map`.

bjorn3:
I think it would be fine, but don't know enough about parking-lot's internals to judge.

Amanieu:
I'm very much in favor of this change. In fact I would like to make this a breaking change and directly change `MappedMutexGuard` instead of adding `OwnedMappedMutexGuard`.

Could you also extend this to support other lock types (rwlock, reentrant mutex, etc)?

yume-chan:
I also want this. My mapper `f` returns a `Box<T>` so I need the guard to own the data. However, instead of `Mutex` I'm using `RwLock`.

Amanieu:
@Vurich Are you still working on this PR?

Vurich:
I’m sorry, I was working on this as part of a project at work but we went with a solution that no longer needed this change anyway, so I forgot about it. I can finish up this PR and we can get it merged.

Amanieu:
Thanks, that would be great!

Amanieu:
Hi, are you still updating this PR? Otherwise I can take over, finish it and get it merged.

lock_api/src/mutex.rs
Vurich: I'm trying to work out what the correct behaviour here would actually be - I've tried to work out a way to cause an invalid reference using the `f` closure but as far as I see it the lifetime bounds make it fine. I think just using `mem::forget(s)` after the `match` instead of using `mem::ManuallyDrop` is the correct behaviour.

I would have preferred to use `catch_unwind`/`resume_unwind` to make the control flow more explicit but unfortunately that's not available with `no_std` right now.

lock_api/src/mutex.rs
```diff
@@ -508,13 +508,15 @@ impl<'a, R: RawMutex + 'a, T: ?Sized + 'a> MutexGuard<'a, R, T> {
     /// used as `MutexGuard::map(...)`. A method would interfere with methods of
     /// the same name on the contents of the locked data.
     #[inline]
-    pub fn map<U: ?Sized, F>(s: Self, f: F) -> MappedMutexGuard<'a, R, U>
+    pub fn map<U, F>(s: Self, f: F) -> MappedMutexGuard<'a, R, U>
     where
-        F: FnOnce(&mut T) -> &mut U,
+        F: FnOnce(&'a mut T) -> U,
```
Amanieu, line 513: This is almost certainly wrong: `'a` is the lifetime of the `Mutex` object, which can be `'static`.
```suggestion
        F: FnOnce(&mut T) -> U,
```

lock_api/src/mutex.rs
```diff
@@ -532,16 +534,22 @@ impl<'a, R: RawMutex + 'a, T: ?Sized + 'a> MutexGuard<'a, R, T> {
     /// used as `MutexGuard::try_map(...)`. A method would interfere with methods of
     /// the same name on the contents of the locked data.
     #[inline]
-    pub fn try_map<U: ?Sized, F>(s: Self, f: F) -> Result<MappedMutexGuard<'a, R, U>, Self>
+    pub fn try_map<U, F>(s: Self, f: F) -> Result<MappedMutexGuard<'a, R, U>, Self>
     where
-        F: FnOnce(&mut T) -> Option<&mut U>,
+        F: FnOnce(&'a mut T) -> Option<U>,
```
Amanieu, line 539: Same here.

lock_api/src/mutex.rs
```diff
@@ -780,10 +788,12 @@ impl<R: RawMutex, T: ?Sized> Drop for ArcMutexGuard<R, T> {
 /// could introduce soundness issues if the locked object is modified by another
 /// thread.
 #[must_use = "if unused the Mutex will immediately unlock"]
-pub struct MappedMutexGuard<'a, R: RawMutex, T: ?Sized> {
+pub struct MappedMutexGuard<'a, R: RawMutex, T: ?Sized + 'a> {
     raw: &'a R,
-    data: *mut T,
-    marker: PhantomData<&'a mut T>,
+    // We use `&'a mut` to make this type invariant over `'a`
+    marker: PhantomData<&'a mut ()>,
```
Amanieu, line 794: Does it actually need to be invariant here? I think we can safely remove this.

lock_api/src/mutex.rs
```diff
@@ -780,10 +788,12 @@ impl<R: RawMutex, T: ?Sized> Drop for ArcMutexGuard<R, T> {
 /// could introduce soundness issues if the locked object is modified by another
 /// thread.
 #[must_use = "if unused the Mutex will immediately unlock"]
-pub struct MappedMutexGuard<'a, R: RawMutex, T: ?Sized> {
+pub struct MappedMutexGuard<'a, R: RawMutex, T: ?Sized + 'a> {
```
Amanieu, line 791: I don't think `T` has to live as long as `'a`, we can remove the `'a` bound on `T`.

Only `R` needs the bound because we have a `&'a R`.

lock_api/src/mutex.rs
```diff
@@ -805,13 +815,19 @@ impl<'a, R: RawMutex + 'a, T: ?Sized + 'a> MappedMutexGuard<'a, R, T> {
     /// used as `MappedMutexGuard::map(...)`. A method would interfere with methods of
     /// the same name on the contents of the locked data.
     #[inline]
-    pub fn map<U: ?Sized, F>(s: Self, f: F) -> MappedMutexGuard<'a, R, U>
+    pub fn map<U, F>(mut s: Self, f: F) -> MappedMutexGuard<'a, R, U>
     where
-        F: FnOnce(&mut T) -> &mut U,
+        F: FnOnce(T) -> U,
     {
+        use core::ptr;
```
Amanieu, line 822: Move this to the top of the file.

lock_api/src/mutex.rs
Amanieu: There is a potential double-drop here: if `f` panics then `data` will be dropped again despite having already been read from `s.data`.

lock_api/src/mutex.rs
```diff
@@ -805,13 +815,19 @@ impl<'a, R: RawMutex + 'a, T: ?Sized + 'a> MappedMutexGuard<'a, R, T> {
...
+
         let raw = s.raw;
-        let data = f(unsafe { &mut *s.data });
+
+        let data = unsafe { ptr::read(&mut s.data) };
```
Amanieu, line 826:
```suggestion
        let data = unsafe { ptr::read(&s.data) };
```

lock_api/src/mutex.rs
Vurich: `'a` is the lifetime of the `MutexGuard`'s lock of the `Mutex`, which isn't the same and to my knowledge it is valid to get an `&'a mut T` from a `MutexGuard<'a, _, T>` since for the duration of `'a` the only reachable path to the `T` is via the `MutexGuard`, and the `MutexGuard` is consumed in this method. Can you come up with an example of a piece of invalid code that would be enabled by this?

lock_api/src/mutex.rs
Vurich: I don't think that is true, `forget` doesn't invalidate anything. For example, `Box::leak` leaks the value and returns a borrow of any length.

lock_api/src/mutex.rs
Vurich: It definitely doesn't need the `'a` bound on the struct itself, and I think you're right that it isn't needed elsewhere either.

lock_api/src/mutex.rs
Vurich: I kept it invariant because that was how it was in the original code, and it's the most conservative option. Without it this type would be variant, which is probably correct but there's a chance of it being subtly broken.

lock_api/src/mutex.rs
Vurich: So I gave it some thought and I think I'll remove the lifetime, it's the more-conservative and more-obviously-correct option and it still allows for 99% of valid usecases, most importantly `MutexGuard::map(foo, |a| &mut a.some_field)`.

Amanieu:
```rust
    #[test]
    fn test_map() {
        let mut outer = &mut 0;
        static M: Mutex<i32> = Mutex::const_new(super::RawMutex::INIT, 0);
        let guard = super::MutexGuard::map(M.lock(), |inner: &'static mut i32| {
            outer = inner;
        });
        drop(guard);
        *outer = 1;
    }
```

Amanieu: `'a` is the lifetime of the `Mutex`. The lifetime of the lock is the lifetime of `self`.

Amanieu: Hmm after thinking about it some more, I think the change proposed in this PR can't possibly work with Rust's current lifetime system. Supporting this would require some form of support for higher-kinded types in the language.

Vurich: How do you mean? I'm not sure I understand what the issue is once you disallow `'a` access to the data.

Amanieu: Well if you make the change that I suggested of removing the `'a` from the `FnOnce` then this code stops working:
```rust
    #[test]
    fn test_map() {
        let m = Mutex::new((0, 0));
        let guard = super::MutexGuard::map(m.lock(), |inner| &mut inner.0);
    }
```

```
error: lifetime may not live long enough
   --> src/mutex.rs:317:62
    |
317 |         let guard = super::MutexGuard::map(m.lock(), |inner| &mut inner.0);
    |                                                       ------ ^^^^^^^^^^^^ returning this value requires that `'1` must outlive `'2`
    |                                                       |    |
    |                                                       |    return type of closure is &'2 mut i32
    |                                                       has type `&'1 mut (i32, i32)`
```

Vurich: So I thought this could be fixed by adding a `'a` bound to `F`, but that doesn't work either:

```rust
#[test]
fn test_map() {
    use crate::{lock_api::RawMutex, Mutex, MutexGuard};

    const FOO: usize = 0;

    static OUTER: Mutex<&usize> = Mutex::const_new(RawMutex::INIT, &FOO);
    static M: Mutex<usize> = Mutex::const_new(RawMutex::INIT, 0);

    let guard = MutexGuard::map(M.lock(), |inner: &'static mut usize| {
        *OUTER.lock() = inner;
    });
    drop(guard);

    let outer: &usize = &*OUTER.lock();

    assert_eq!(*outer, 0);

    *M.lock() = 1;

    assert_eq!(*outer, 1);
}
```

Vurich: The issue is that we want the `U` that's returned from the function to be dependent on the input lifetime - we want to give the function a `&'a mut T` so that it's possible for it to return a field, but we _also_ don't want to let it rely on getting an `&'a mut T`. For references that's trivial - you can do `for<'any> F: FnOnce(&'any mut T) -> &'any mut U`, and in fact that's what the current code does - but for generic types that's not possible.

lock_api/src/mutex.rs
bjorn3: You are right. This doesn't actually invalidate the borrow because `s.mutex` is a reference.

lock_api/src/mutex.rs
Vurich: Another issue is that it's not possible to restrict a type's lifetime generically - you can do `for<'any> F: FnOnce(&'any mut T) -> &'any mut U`, and then essentially convert your `&'a mut T` to a `for<'any> &'any mut T`, but you can't do the same if the `&'a mut T` is "hidden" by being some generic parameter. There's no equivalent of `for<'any> F: FnOnce(&'any mut T) -> &'any mut U` for `F: FnOnce(T) -> U`.

Vurich: This approach appears to work and disallows the code we want to disallow while allowing the code we want to allow, but the ergonomics are quite bad (I'll explain what the problem is below):

```rust
pub trait FnOnceShim<'a, T: 'a> {
    type Output: 'a;

    fn call(self, input: T) -> Self::Output;
}

impl<'a, F, In, Out> FnOnceShim<'a, In> for F
where
    F: FnOnce(In) -> Out,
    In: 'a,
    Out: 'a,
{
    type Output = Out;

    fn call(self, input: In) -> Self::Output {
        self(input)
    }
}

impl<'a, R: RawMutex + 'a, T: ?Sized + 'a> MutexGuard<'a, R, T> {
    // ...

    pub fn map<F>(
        s: Self,
        f: F,
    ) -> MappedMutexGuard<'a, R, <F as FnOnceShim<'a, &'a mut T>>::Output>
    where
        for<'any> F: FnOnceShim<'any, &'any mut T>,
    {
        let raw = &s.mutex.raw;
        let data = unsafe { &mut *s.mutex.data.get() };
        mem::forget(s);

        let data = f.call(data);
        MappedMutexGuard {
            raw,
            data,
            marker: PhantomData,
        }
    }
    
    // ...
}

impl<'a, R: RawMutex + 'a, T: 'a> MappedMutexGuard<'a, R, T> {
    // ...
    
    pub fn map<F>(s: Self, f: F) -> MappedMutexGuard<'a, R, <F as FnOnceShim<'a, T>>::Output>
    where
        for<'any> F: FnOnceShim<'any, T>,
    {
        let (data, raw) = {
            let s = mem::ManuallyDrop::new(s);
            (unsafe { ptr::read(&s.data) }, s.raw)
        };

        // `panic::catch_unwind` isn't available in `core`, so we use a dummy guard to unlock
        // the mutex in case of unwind.
        let lock_guard: MappedMutexGuard<'a, R, ()> = MappedMutexGuard {
            raw,
            data: (),
            marker: PhantomData,
        };

        let data = f.call(data);

        mem::forget(lock_guard);

        MappedMutexGuard {
            raw,
            data,
            marker: PhantomData,
        }
    }

    // ...
}
```

The issue is that Rust will no longer automatically convert the closure to the most-general version of it like it will when being passed as an `F: FnOnce(..) -> ..`. Therefore, you need to do something like this:

```rust
MutexGuard::map(M.lock(), { |inner| inner } as fn(&mut usize) -> &mut usize)
```

Amanieu: Would it help if you changed the bounds to contain both `FnOnce` and `FnOnceShim`? Even though it is redundant, it may help with type inference.

Amanieu:
Unfortunately the need to cast the closure to a `fn` is a dealbreaker: it makes the interface unergonomic and doesn't work with closures that capture state.

Vurich:
> Unfortunately the need to cast the closure to a `fn` is a dealbreaker: it makes the interface unergonomic and doesn't work with closures that capture state.

You don't need to cast to `fn`, you just need to force Rust to infer the most-general lifetime for the argument. There are a couple of ways to do this, but they all work in essentially the same way:

```rust
// Option 1
fn annotate<F, T, U>(f: F) -> F
where
    F: FnOnce(&mut T) -> &mut U,
{
    f
}

let func = annotate(|foo: &mut (_, _)| &mut foo.0);

// Option 2
fn create_mapping_func<A, B>() -> impl FnOnce(&mut (A, B)) -> A {
    |foo| &mut foo.0
}

// Option 3
let func: &mut dyn FnMut(&mut (_, _)) -> &mut _ = &mut |foo| &mut foo.0;
```

All of these are pretty bad though, sadly. Rust has a long-standing issue open to fix this, but it has yet to be addressed https://github.com/rust-lang/rust/issues/70263

lock_api/src/mutex.rs
a1phyr: Actually this made `T` invariant, not `'a` (see https://doc.rust-lang.org/nomicon/subtyping.html#variance).

The proper way to get invariance on `'a` is  `PhantomData<fn(&'a ()) -> &'a ()>`

PR #291, opened 2021-07-29
Add arc_lock feature, with guards that can be accessed from inside of Arc

### LINKED ISSUE #273: Add arc-based mutex locks to the `lock_api` crate
At the moment, mutex locks are implemented roughly as follows:

```
pub struct MutexGuard<'a, T> {
    inner: &'a Mutex<T>,
}
```

This is fine for cases where the `Mutex` is standalone. However, consider the common use pattern for a Mutex where it is put into an `Arc` to give the `Arc` a measure of safe interior mutability. Since an `Arc<T>`, when boiled down, can act as `&T`, one could create a pattern as follows:

```
pub struct MutexGuardArc<T> {
    inner: Arc<Mutex<T>>,
}

impl<T> Mutex<T> {
    pub fn lock_arc(this: &Arc<Self>) -> MutexGuardArc<T> {
        MutexGuardArc {
            inner: this.clone(),
        }
    }
}
```

The advantage of this is that this pattern elides the use of a lifetime which, for me personally, comes in handy when I'm using an `Arc<Mutex>` in cases where using lifetimes would unnecessarily complicate things. [`async_lock::Mutex`](https://docs.rs/async-lock/2.3.0/async_lock/struct.MutexGuardArc.html) provides this pattern already, and it would be nice if `lock_api` could provide that pattern as well.

I am willing to implement this myself if you approve of it.

notgull:
Resolves #273.

This adds guards for `Mutex` and `RwLock` that allow it to be used from inside of an `Arc`, without lifetime hurdles.

### DIFF
--- a/lock_api/src/mutex.rs
+++ b/lock_api/src/mutex.rs
@@ -13,6 +13,10 @@ use core::ops::{Deref, DerefMut};
 
 #[cfg(feature = "arc_lock")]
 use alloc::sync::Arc;
+#[cfg(feature = "arc_lock")]
+use core::mem::ManuallyDrop;
+#[cfg(feature = "arc_lock")]
+use core::ptr;
 
 #[cfg(feature = "owning_ref")]
 use owning_ref::StableAddress;
@@ -295,9 +299,9 @@ impl<R: RawMutex, T: ?Sized> Mutex<R, T> {
     /// The lock needs to be held for the behavior of this function to be defined.
     #[cfg(feature = "arc_lock")]
     #[inline]
-    unsafe fn guard_arc(this: &Arc<Self>) -> ArcMutexGuard<R, T> {
+    unsafe fn guard_arc(self: &Arc<Self>) -> ArcMutexGuard<R, T> {
         ArcMutexGuard {
-            mutex: this.clone(),
+            mutex: self.clone(),
             marker: PhantomData,
         }
     }
@@ -308,10 +312,10 @@ impl<R: RawMutex, T: ?Sized> Mutex<R, T> {
     /// and the resulting mutex guard has no lifetime requirements.
     #[cfg(feature = "arc_lock")]
     #[inline]
-    pub fn lock_arc(this: &Arc<Self>) -> ArcMutexGuard<R, T> {
-        this.raw.lock();
+    pub fn lock_arc(self: &Arc<Self>) -> ArcMutexGuard<R, T> {
+        self.raw.lock();
         // SAFETY: the locking guarantee is upheld
-        unsafe { Self::guard_arc(this) }
+        unsafe { self.guard_arc() }
     }
 
     /// Attempts to acquire a lock through an `Arc`.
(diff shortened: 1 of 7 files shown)


lock_api/src/mutex.rs
```diff
@@ -336,6 +378,39 @@ impl<R: RawMutexTimed, T: ?Sized> Mutex<R, T> {
...
         }
     }
+
+    /// Attempts to acquire this lock through an `Arc` until a timeout is reached.
+    ///
+    /// This method is similar to the `try_lock_for` method; however, it requires the `Mutex` to be inside of an
+    /// `Arc` and the resulting mutex guard has no lifetime requirements.
+    #[cfg(feature = "arc_lock")]
+    #[inline]
+    pub fn try_lock_for_arc(this: &Arc<Self>, timeout: R::Duration) -> Option<ArcMutexGuard<R, T>> {
```
Amanieu, line 388: `try_lock_arc_for` sounds better here.

lock_api/src/mutex.rs
```diff
@@ -336,6 +378,39 @@ impl<R: RawMutexTimed, T: ?Sized> Mutex<R, T> {
...
+        }
+    }
+
+    /// Attempts to acquire this lock through an `Arc` until a timeout is reached.
+    ///
+    /// This method is similar to the `try_lock_until` method; however, it requires the `Mutex` to be inside of
+    /// an `Arc` and the resulting mutex guard has no lifetime requirements.
+    #[cfg(feature = "arc_lock")]
+    #[inline]
+    pub fn try_lock_until_arc(
```
Amanieu, line 403: Same here.

lock_api/src/mutex.rs
```diff
@@ -583,6 +658,112 @@ impl<'a, R: RawMutex + 'a, T: fmt::Display + ?Sized + 'a> fmt::Display for Mutex
 #[cfg(feature = "owning_ref")]
 unsafe impl<'a, R: RawMutex + 'a, T: ?Sized + 'a> StableAddress for MutexGuard<'a, R, T> {}
 
+/// An RAII mutex guard returned by the `Arc` locking operations on `Mutex`. This is similar to the `MutexGuard`
```
Amanieu, line 661: Split the first sentence into a separate paragraph. This summary is shown in the module documentation index.

lock_api/src/mutex.rs
```diff
@@ -286,6 +289,45 @@ impl<R: RawMutex, T: ?Sized> Mutex<R, T> {
...
+        }
+    }
+
+    /// Acquires a lock through an `Arc`.
+    ///
+    /// This method is similar to the `lock` method; however, it requires the `Mutex` to be inside of an `Arc`
+    /// and the resulting mutex guard has no lifetime requirements.
+    #[cfg(feature = "arc_lock")]
+    #[inline]
+    pub fn lock_arc(this: &Arc<Self>) -> ArcMutexGuard<R, T> {
```
Amanieu, line 311:
```suggestion
    pub fn lock_arc(self: &Arc<Self>) -> ArcMutexGuard<R, T> {
```
You can use `self` here which allows `lock_arc` to be called with dot syntax.

Amanieu (commented):
Nice work!

Could you extend this to `ReentrantMutex` and `RwLock` as well?

Amanieu:
Great work! Can you update the CI script so that the `arc_lock` feature is tested?

notgull:
It doesn't seem like you can use `Arc` in `self` without an MSRV bump, unfortunately.

Amanieu:
It's fine I'll just bump the MSRV for the next release.

PR #302, opened 2021-10-31
Disable eventual fairness on wasm32-unknown-unknown

### LINKED ISSUE #269: parking_lot_core 0.8.1 wasm link issue
Starting with parking_lot_core 0.8.1, when compiling to wasm there is now a depdendency on "env".

This means one can no longer use parking_lot from within a wasm runtime, such as [`wasmtime`](https://github.com/bytecodealliance/wasmtime).

When compiling for the web, one can address this problem with `parking_lot = { version = "0.11", features = ["wasm-bindgen"] }`, but this does not work for non-web wasm runtimes.

## Reproduce

`Cargo.toml`:

``` toml
[package]
name = "parking_lot_test"
version = "0.1.0"

[dependencies]
parking_lot = "0.11"
parking_lot_core = "=0.8.1"
```

`main.rs`:

``` rust
fn main() {
    let m = parking_lot::Mutex::new(());
    let _lock = m.lock();
}
```

Using [`wasm2wat`](https://github.com/WebAssembly/wabt):

```
§ cargo build --target "wasm32-unknown-unknown"
§ wasm2wat target/wasm32-unknown-unknown/debug/parking_lot_test.wasm | rg import
> (import "env" "now" (func $now (type 7)))
```

(with `parking_lot_core = "=0.8.0"` there is no `import "env"`)

Amanieu:
std::time::Instant panics on this target and the external instant crate
adds undesirable external dependencies to the wasm module.

Fixes #269

### DIFF
--- a/core/src/parking_lot.rs
+++ b/core/src/parking_lot.rs
@@ -18,7 +18,7 @@ use std::time::{Duration, Instant};
 // Don't use Instant on wasm32-unknown-unknown, it just panics.
 cfg_if::cfg_if! {
     if #[cfg(all(
-        target_arch = "wasm32",
+        target_family = "wasm",
         target_os = "unknown",
         target_vendor = "unknown"
     ))] {
--- a/core/src/parking_lot.rs
+++ b/core/src/parking_lot.rs
@@ -12,9 +12,33 @@ use core::{
     ptr,
     sync::atomic::{AtomicPtr, AtomicUsize, Ordering},
 };
-use instant::Instant;
 use smallvec::SmallVec;
-use std::time::Duration;
+use std::time::{Duration, Instant};
+
+// Don't use Instant on wasm32-unknown-unknown, it just panics.
+cfg_if::cfg_if! {
+    if #[cfg(all(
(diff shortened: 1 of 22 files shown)


core/src/parking_lot.rs
```diff
@@ -104,14 +121,21 @@ impl FairTimeout {
...
-        let now = Instant::now();
-        if now > self.timeout {
-            // Time between 0 and 1ms.
-            let nanos = self.gen_u32() % 1_000_000;
-            self.timeout = now + Duration::new(0, nanos);
-            true
-        } else {
-            false
+        #[cfg(not(all(
+            target_arch = "wasm32",
```
CryZe, line 125:
```suggestion
            target_family = "wasm",
```

CryZe: Though this is only for up to date compilers, but it's future proof towards wasm64.

d3lm:
FWIW, I think it only adds undesirable external dependencies if your host is not a browser. If you target the browser you can configure it and enable features and it compiles just fine.

Issue #293, opened 2021-08-10
Add is_locked_exclusive to RwLock


topecongiro:
It would be great if you could add `is_locked_exclusive` (or whatever the name is) to `RwLock` API. The method shall return `true` if the lock is held exclusively and `false` otherwise, ideally without modifying the underlying lock data structure.

The method will be useful when constructing certain algorithms (e.g., optimistic latching). Currently, `parking_lot` only exposes `is_locked` which returns true against both shared/exclusive, or `try_write` which will modify the lock data structure.

declanvk:
This should have been closed by #303 , @topecongiro could you check out the PR and see if it works for you?

topecongiro:
@declanvk Thank you for the PR, and sorry for the late reply.

Regarding implementing HybridLocking in the paper, I think that we need to use `Ordering::Acquire` in `is_locked` and `is_locked_exclusive`, or otherwise, there is a chance that the instruction reading the version gets reordered before `is_locked_exclusive`.

PR #303, opened 2021-11-04
Add a rwlock method to report if lock is held exclusively

declanvk:
Closes issue #293.

I also ran into the same issue posed in issue #293. Specifically, it was
for implementing an optimistic hybrid lock as described in [this paper][scalable-latches]
in section 2.3, listing #2. 

[scalable-latches]: https://db.in.tum.de/~giceva/papers/damon_latches.pdf?lang=de

### DIFF
--- a/lock_api/src/rwlock.rs
+++ b/lock_api/src/rwlock.rs
@@ -86,7 +86,16 @@ pub unsafe trait RawRwLock {
     }
 
     /// Check if this `RwLock` is currently exclusively locked.
-    fn is_locked_exclusive(&self) -> bool;
+    fn is_locked_exclusive(&self) -> bool {
+        let acquired_lock = self.try_lock_shared();
+        if acquired_lock {
+            // Safety: A shared lock was successfully acquired above.
+            unsafe {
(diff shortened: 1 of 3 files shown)


Amanieu:
I think you could add a default implementation in the trait which attempts to get a shared lock. This should only fail if there is an active exclusive lock. The default implementation is needed to maintain backwards compatibility with existing implementations of this trait.

declanvk:
> I think you could add a default implementation in the trait which attempts to get a shared lock. This should only fail if there is an active exclusive lock. The default implementation is needed to maintain backwards compatibility with existing implementations of this trait.

I'll add that to the trait definition. To test for an exclusive lock, would it look something like:
```rust
    fn is_locked_exclusive(&self) -> bool {
        let acquired_lock = self.try_lock_shared();
        if acquired_lock {
            // Safety: A shared lock was successfully acquired above.
            unsafe {
                self.unlock_shared();
            }
        }
        !acquired_lock
    }
```
to parallel the existing `is_locked` default implementation?

Amanieu:
That's correct.

PR #304, opened 2021-11-11
Switch from llvm_asm!() to new asm!() macro

coolreader18:
To be honest I'm not the most familiar with inline assembly, but afaict the new version is correct. The assembly output is identical, at least, checking `RawRwLock::bump_shared_slow` (which calls both elision impls) using `cargo asm`.

This should be ready for use when `asm!()` is stabilized, cf (well, your own stabilization report I guess) https://github.com/rust-lang/rust/issues/72016#issuecomment-964186287

### DIFF
--- a/src/elision.rs
+++ b/src/elision.rs
@@ -56,6 +56,16 @@ impl AtomicElisionExt for AtomicUsize {
     fn elision_compare_exchange_acquire(&self, current: usize, new: usize) -> Result<usize, usize> {
         unsafe {
             let prev: usize;
+            #[cfg(target_pointer_width = "32")]
+            asm!(
+                "xacquire",
+                "lock",
+                "cmpxchg [{:e}], {:e}",
+                in(reg) self,
+                in(reg) new,
+                inout("eax") current => prev,
+            );
+            #[cfg(target_pointer_width = "64")]
             asm!(
                 "xacquire",
                 "lock",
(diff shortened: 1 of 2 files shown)


Amanieu:
This code is actually incorrect on the x32 ABI: you need to force the use of the `:e` modifier for both registers so that a 32-bit cmpxchg is used instead of a 64-bit one.

coolreader18:
Not that I'd know more about this than you, but are you sure? The unstable book looks like it implies that eax will be chosen on x86 32-bit (https://doc.rust-lang.org/unstable-book/library-features/asm.html)

Architecture | Register class | Modifier | Example output | LLVM modifier
-- | -- | -- | -- | --
x86-32 | reg | None | eax | k
x86-64 | reg | None | rax | q

Amanieu:
x32 is not the same as x86. It's the 32-bit ABI for x86_64 where it uses the x86_64 ISA but with 32-bit pointers. You can try it yourself with the `x86_64-unknown-linux-gnux32` target.

src/elision.rs
```diff
@@ -87,29 +82,17 @@ impl AtomicElisionExt for AtomicUsize {
...
             let prev: usize;
-            llvm_asm!("xrelease; lock; xaddq $2, $1"
-                      : "=r" (prev), "+*m" (self)
-                      : "0" (val.wrapping_neg())
-                      : "memory"
-                      : "volatile");
+            asm!(
+                "xrelease",
+                "lock",
+                "xadd [{}], {}",
```
Amanieu, line 92: You also need `:e` here for 32-bit pointer size.

PR #309, opened 2021-12-19
Fix build on latest nightly

Amanieu:
asm! is now stable.

### DIFF
--- a/src/elision.rs
+++ b/src/elision.rs
@@ -5,6 +5,8 @@
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
+#[cfg(all(feature = "nightly", any(target_arch = "x86", target_arch = "x86_64")))]
+use std::arch::asm;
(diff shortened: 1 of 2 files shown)


src/elision.rs
```diff
@@ -5,6 +5,8 @@
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
+#[cfg(all(feature = "nightly", any(target_arch = "x86", target_arch = "x86_64")))]
```
bjorn3, line 8: Maybe rename the feature to "hardware-lock-elision" now or in a future release?

Amanieu: In a future version I will just have this always enabled by default.

bjorn3: Don't some processors have a bug causing "unpredictable system behavior" (aka crashes or worse) when using hardware lock elision?

Amanieu: That's actually a good point, I wasn't aware that HLE was broken on all currently available CPUs.

PR #311, opened 2022-01-08
Adopt windows-rs

clemenswasser:
Change from using the unofficial winapi bindings, to using the official windows bindings from Microsoft.
I have also a PR open at winit: https://github.com/rust-windowing/winit/pull/2057
They suggested that parking_lot (dependency of winit) should also be switched over to windows-rs.
For this port I am using the more lightweight and lower level windows-sys (which is closer to winapi).
For me build times are faster with windows-rs 🎉 
Before(winapi 1.19s, total time 2.2s):
winapi_parking_lot
After(windows-sys 0.94s, total time 1.7s):
windows_parking_lot

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -23,8 +23,12 @@ libc = "0.2.95"
 redox_syscall = "0.2.8"
 
 [target.'cfg(windows)'.dependencies]
-winapi = { version = "0.3.9", features = ["winnt", "ntstatus", "minwindef",
-    "winerror", "winbase", "errhandlingapi", "handleapi"] }
+windows-sys = { version = "0.29", features = [
+    "Win32_Foundation",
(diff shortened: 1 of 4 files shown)


bjorn3:
Several other crates also compile faster in the second compile. Are you sure this isn't the cpu changing frequency?

clemenswasser:
> Several other crates also compile faster in the second compile. Are you sure this isn't the cpu changing frequency?

It is deterministically faster.
When I build it multiple times, it's always winit ~2s and windows-rs ~1.7s.

Amanieu:
Thanks!

madsmtm:
Note that this might be an MSRV bump (at least when the `windows` crate had an MSRV, see https://github.com/microsoft/windows-rs/issues/1426#issuecomment-1011795357)

PR #314, opened 2022-01-28
Use core::hint::spin_loop instead of the deprecated spin_loop_hint.

### LINKED ISSUE #295: CI is running an extremely old version of Rust
While working on https://github.com/Amanieu/parking_lot/pull/290, I fixed a deprecation warning (`std::sync::atomic::spin_loop_hint` is now `std::hint::spin_loop`). However, this caused errors on CI because it is running Rust 1.41, whereas the latest stable version is 1.54. This might be intentional, as `parking_lot` is a reasonably fundamental crate in the Rust ecosystem and so should support older versions, but 1.41 is now over a year and a half old, `std::hint::spin_loop` has been stable for almost a year and `spin_loop_hint` has been deprecated for almost as long as that.

Amanieu:
Bumps MSRV to 1.49.0

Fixes #295

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -16,11 +16,11 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu, macos, windows]
-        channel: [1.41.0, stable, beta, nightly]
+        channel: [1.49.0, stable, beta, nightly]
         feature: [arc_lock, serde, deadlock_detection]
         exclude: 
           - feature: deadlock_detection
-            channel: '1.41.0'
+            channel: '1.49.0'
         include:
           - channel: nightly
             feature: nightly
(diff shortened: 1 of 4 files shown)


.github/workflows/rust.yml
```diff
@@ -16,11 +16,11 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu, macos, windows]
-        channel: [1.41.0, stable, beta, nightly]
+        channel: [1.49.0, stable, beta, nightly]
         feature: [arc_lock, serde, deadlock_detection]
         exclude: 
           - feature: deadlock_detection
-            channel: '1.41.0'
+            channel: '1.48.0'
```
CryZe, line 23: Was this meant to be 1.49?

Amanieu: Oops, nice catch.

PR #325, opened 2022-03-10
Update to stabilized const_fn_trait_bound

taiki-e:
const_fn_trait_bound has been stabilized on Rust 1.61 (currently nightly, nightly-2022-03-08+): https://github.com/rust-lang/rust/pull/93827

This PR removes `#![feature(const_fn_trait_bound)]` and makes constructors for lock types `const fn` on Rust 1.61+ using autocfg.


lock_api/build.rs
```diff
@@ -0,0 +1,16 @@
+fn main() {
```
bjorn3, line 1: Adding a build script will make building a bit slower as the build script has to be compiled and the main crate has to wait on running the build script.

taiki-e: Yeah. An alternative approach without such drawbacks is to use the options feature. If the maintainer prefers that approach, I'm happy to switch to using it.

Amanieu: I personally don't mind this: the build script has no dependencies and can be compiled in parallel with other crates, so it is usually not noticable.

PR #330, opened 2022-03-18
Use Ordering::Acquire for is_locked predicates

topecongiro:
In algorithms such as optimistic latch coupling, reordering instructions before and after is_locked() or is_locked_exclusive() could cause a problem.

In the current implementation, we use the relaxed memory ordering for reading the state value. This allows instructions after is_locked()/is_locked_exclusive() to be reordered before them. This will break the aforementioned optimistic latch coupling.

### DIFF
--- a/src/raw_rwlock.rs
+++ b/src/raw_rwlock.rs
@@ -140,13 +140,13 @@ unsafe impl lock_api::RawRwLock for RawRwLock {
 
     #[inline]
     fn is_locked(&self) -> bool {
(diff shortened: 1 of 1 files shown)


Amanieu:
I feel that it should be your responsibility to insert a `fence(Ordering::Acquire)` after checking `is_locked` if your code requires it.

topecongiro:
I think you are certainly right now that I think about it. I will close this as my use case can be covered by manually inserting `fence` when necessary. Thank you for the review!

PR #339, opened 2022-04-27
deps: update to windows-sys 0.36

poliorcetics:


poliorcetics:
This is a separate PR from #337 because some people will probably want a dep on 0.35 and others a dep on 0.36, to avoid duplicate dependencies in their tree.

As such, #337 should be merged first, a version released, then this one and another version released

poliorcetics:
See also https://github.com/microsoft/windows-rs/issues/1720

kennykerr:
tokio/mio just took a dependency on 0.36 so that may be a good version to rally around for the time being.

https://github.com/tokio-rs/mio/pull/1569

Amanieu:
I'm just going to merge this one and make a single release. This version churn is quite annoying, and if it continues then I'm considering just switching back to winapi.

kennykerr:
Sorry for the annoyance. I'll probably exclude `windows-sys` from the regular 3-week release cadence that we've been on to allow the current version (0.36) to establish itself.

PR #340, opened 2022-04-30
Update dependencies

jelmansouri:
We're using cargo deny on our project, and try to keep dependencies consistent. Can a path be released with the following update please!


jelmansouri:
there is another PR :) https://github.com/Amanieu/parking_lot/pull/339

PR #341, opened 2022-05-03
Shrink unsafe blocks 

peamaeq:
The scope of the unsafe blocks can be appropriately reduced

### DIFF
--- a/src/condvar.rs
+++ b/src/condvar.rs
@@ -136,11 +136,10 @@ impl Condvar {
 
     #[cold]
     fn notify_one_slow(&self, mutex: *mut RawMutex) -> bool {
-        unsafe {
-            // Unpark one thread and requeue the rest onto the mutex
-            let from = self as *const _ as usize;
-            let to = mutex as usize;
-            let validate = || {
+        // Unpark one thread and requeue the rest onto the mutex
+        let from = self as *const _ as usize;
+        let to = mutex as usize;
+            let validate = unsafe { || {
                 // Make sure that our atomic state still points to the same
                 // mutex. If not then it means that all threads on the current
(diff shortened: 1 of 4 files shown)


src/condvar.rs
```diff
@@ -197,11 +195,10 @@ impl Condvar {
 
     #[cold]
     fn notify_all_slow(&self, mutex: *mut RawMutex) -> usize {
-        unsafe {
             // Unpark one thread and requeue the rest onto the mutex
             let from = self as *const _ as usize;
             let to = mutex as usize;
-            let validate = || {
+            let validate = unsafe { || {
```
Amanieu, line 201: Can this `unsafe` be moved inside the closure?

src/condvar.rs
```diff
@@ -317,10 +312,10 @@ impl Condvar {
                     }
                     true
                 };
-                let before_sleep = || {
+                let before_sleep = unsafe { || {
```
Amanieu, line 315: Same here.

src/condvar.rs
```diff
@@ -136,11 +136,10 @@ impl Condvar {
...
     fn notify_one_slow(&self, mutex: *mut RawMutex) -> bool {
-        unsafe {
-            // Unpark one thread and requeue the rest onto the mutex
-            let from = self as *const _ as usize;
-            let to = mutex as usize;
-            let validate = || {
+        // Unpark one thread and requeue the rest onto the mutex
+        let from = self as *const _ as usize;
+        let to = mutex as usize;
+            let validate = unsafe { || {
```
Amanieu, line 142: And here.

src/condvar.rs
peamaeq: Sorry for the long reply, very good suggestion

Issue #329, opened 2022-03-13
Condvar::wait_while and Condvar::wait_timeout_while convenience methods


bryanhitc:
The standard library has  `Convar::wait_while` and `Convar::wait_timeout_while` methods that take a `condition: impl Fn(&mut T) -> bool`. These convenience methods are great since the following pattern is relatively common:

```rs
while condition(&mut *lock_guard) {
  cv.wait(lock_guard);
}

// condition is now false
```

I think it'd be great to have these convenience methods built-in, but I can also see an argument being made for not wanting to clutter the public API. Thoughts?

I'm more than happy to submit the PR if this functionality is desired.

Amanieu:
Sure! I'd be happy to accept a PR for this.

PR #343, opened 2022-05-29
Add Condvar::wait_while convenience methods

bryanhitc:
Addresses #329.

Let me know what your thoughts are. I'm not sure if `WaitWhileResult` is useful, but I included it anyway since it helps with the tests.

### DIFF
--- a/src/condvar.rs
+++ b/src/condvar.rs
@@ -411,7 +411,7 @@ impl Condvar {
         let mut timeout_result = WaitTimeoutResult(false);
 
         while !timeout_result.timed_out() && condition(mutex_guard.deref_mut()) {
-            result.0 += 1;
+            result.0 = result.0.saturating_add(1);
             timeout_result =
                 self.wait_until_internal(unsafe { MutexGuard::mutex(mutex_guard).raw() }, timeout);
         }
--- a/src/condvar.rs
+++ b/src/condvar.rs
@@ -757,7 +757,7 @@ mod tests {
         let mutex = Arc::new(Mutex::new(0));
         let cv = Arc::new(Condvar::new());
 
-        let num_iters = u32::MAX;
+        let num_iters = 3;
         let condition = |counter: &mut u32| {
             *counter += 1;
             true
@@ -771,9 +771,7 @@ mod tests {
             cv.wait_while_until_internal(&mut mutex_guard, condition, timeout);
 
         assert!(timeout_result.timed_out());
-        // thread should be blocked + woken up multiple times
-        assert!(*mutex_guard > 2);
-        assert!(*mutex_guard < num_iters);
+        assert!(*mutex_guard == num_iters + 1);
 
         // prevent deadlock with notifier
         drop(mutex_guard);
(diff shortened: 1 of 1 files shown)


src/condvar.rs
```diff
@@ -383,6 +395,134 @@ impl Condvar {
         let deadline = util::to_deadline(timeout);
         self.wait_until_internal(unsafe { MutexGuard::mutex(mutex_guard).raw() }, deadline)
     }
+
+    #[inline]
```
bryanhitc, line 399: We may want benchmarks for this, but hinting for inlining seems like a good idea.

I don't think there's a reasonable way to reduce the monomorphization cost like `wait_until_internal` since `condition` is inherently generic.

src/condvar.rs
```diff
@@ -572,6 +714,113 @@ mod tests {
...
+        let mutex = Arc::new(Mutex::new(0));
+        let cv = Arc::new(Condvar::new());
+
+        let condition = |counter: &mut u32| {
+            *counter += 1;
+            true
+        };
+
+        let mut mutex_guard = mutex.lock();
+        let timeout = Some(Instant::now() + Duration::from_millis(50));
```
bryanhitc, line 780: This timeout may result in a flaky test if it's too aggressive, but in practice, I think 50ms should be fine and it doesn't meaningfully impact test execution speed. It would be great if we could inject a virtual clock and then manually advance the time deterministically, but that would require quite a bit of work + complicate the API for questionable benefits.

src/condvar.rs
```diff
@@ -29,6 +30,17 @@ impl WaitTimeoutResult {
...
+/// A type indicating how many times a thread was blocked during wait_while.
+#[derive(Debug, PartialEq, Eq, Copy, Clone)]
+pub struct WaitWhileResult(u32);
+
+impl WaitWhileResult {
+    #[inline]
+    pub fn num_iters(self) -> u32 {
+        self.0
+    }
+}
```
Amanieu, line 42: I don't think this is useful to track. I've never seen any use for this information. It also makes the API unnecessarily complex.

src/condvar.rs
Amanieu: I agree, an `#[inline]` hint would be useful here.

src/condvar.rs
Amanieu: I'd bump the timeout to 500ms just to be safe. The difference in CI times is imperceptible, and tests run in parallel anyways.

src/condvar.rs
bryanhitc: I agree. I removed `WaitWhileResult` and updated the tests.

src/condvar.rs
```diff
@@ -383,6 +384,131 @@ impl Condvar {
...
+    /// # Panics
+    ///
+    /// This function will panic if another thread is waiting on the `Condvar`
+    /// with a different `Mutex` object.
+    #[inline]
+    pub fn wait_while<T, F>(
+        &self,
+        mutex_guard: &mut MutexGuard<'_, T>,
+        condition: F,
+    ) -> WaitTimeoutResult
```
Amanieu, line 429: This shouldn't return a `WaitTimeoutResult` since it cannot time out.

bryanhitc: Oops 😅. Thanks!

PR #349, opened 2022-08-18
Fix `ArcMutexGuard` soundness

### LINKED ISSUE #346: ArcMutexGuard is unsound
The `ArcMutexGuard` type is unsound because it implements `Sync` if `T: Send`, when it should be `Sync` if `T: Sync` (just like the regular MutexGuard)

This code causes a race condition: (tested with miri)
(with `parking_lot` v0.12.1 + feature `arc_lock`)
(no playground link as it requires an extra feature)

```rust
use parking_lot::Mutex;
use std::cell::Cell;
use std::sync::Arc;
use std::thread;

pub fn main() {
    let data = Arc::new(Mutex::new(Cell::new(-1)));

    // this is Sync even though T isn't
    // ArcMutexGuard is Sync if T is Send
    let guard = data.lock_arc();

    thread::scope(|s| {
        for n in 0..10 {
            // closure is move so this moves the ref instead of the whole guard
            // &guard is send even though &cell isnt
            let g = &guard;
            s.spawn(move || {
                let cell: &Cell<i32> = g;
                cell.set(n); // data race
            });
        }
    });

    println!("non-deterministic: {}", guard.get());
}
```

notgull:
Resolves #346

### DIFF
--- a/lock_api/src/mutex.rs
+++ b/lock_api/src/mutex.rs
@@ -681,16 +681,14 @@ unsafe impl<'a, R: RawMutex + 'a, T: ?Sized + 'a> StableAddress for MutexGuard<'
 #[must_use = "if unused the Mutex will immediately unlock"]
 pub struct ArcMutexGuard<R: RawMutex, T: ?Sized> {
     mutex: Arc<Mutex<R, T>>,
-    // n.b. notgull: add a ptr here to prevent the MutexGuard
-    // from being Sync unless it is Sync itself.
-    marker: PhantomData<(R::GuardMarker, *const ())>,
+    marker: PhantomData<*const ()>,
 }
 
-unsafe impl<R: RawMutex, T: Send + Sync + ?Sized> Send for ArcMutexGuard<R, T> where
+unsafe impl<R: RawMutex + Send + Sync, T: Send + Sync + ?Sized> Send for ArcMutexGuard<R, T> where
     R::GuardMarker: Send
 {
 }
-unsafe impl<R: RawMutex, T: Send + Sync + ?Sized> Sync for ArcMutexGuard<R, T> where
+unsafe impl<R: RawMutex + Sync, T: Sync + ?Sized> Sync for ArcMutexGuard<R, T> where
     R::GuardMarker: Sync
 {
 }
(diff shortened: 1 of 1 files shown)


lock_api/src/mutex.rs
```diff
@@ -681,7 +681,18 @@ unsafe impl<'a, R: RawMutex + 'a, T: ?Sized + 'a> StableAddress for MutexGuard<'
...
+    // n.b. notgull: add a ptr here to prevent the MutexGuard
+    // from being Sync unless it is Sync itself.
+    marker: PhantomData<(R::GuardMarker, *const ())>,
+}
+
+unsafe impl<R: RawMutex, T: Send + Sync + ?Sized> Send for ArcMutexGuard<R, T> where
+    R::GuardMarker: Send
+{
+}
+unsafe impl<R: RawMutex, T: Send + Sync + ?Sized> Sync for ArcMutexGuard<R, T> where
```
Amanieu, line 693:
```suggestion
unsafe impl<R: RawMutex, T: Sync + ?Sized> Sync for ArcMutexGuard<R, T> where
```
Same here I think we only require `T: Sync` here.

lock_api/src/mutex.rs
```diff
@@ -681,7 +681,16 @@ unsafe impl<'a, R: RawMutex + 'a, T: ?Sized + 'a> StableAddress for MutexGuard<'
 #[must_use = "if unused the Mutex will immediately unlock"]
 pub struct ArcMutexGuard<R: RawMutex, T: ?Sized> {
     mutex: Arc<Mutex<R, T>>,
-    marker: PhantomData<R::GuardMarker>,
+    marker: PhantomData<*const ()>,
+}
+
+unsafe impl<R: RawMutex + Send + Sync, T: Send + Sync + ?Sized> Send for ArcMutexGuard<R, T> where
```
Amanieu, line 687: I don't think `T: Sync` is necessary here. Everything else looks fine.

Amanieu, line 687: These impls need `#[cfg(feature = "arc_lock")]`.

thalesfragoso: Sorry for digging out this old PR, but...

Can you explain why you think that ? The referenced issue was pointing out exactly that.

If I can send a guard to another thread it means that I can access `&T` from another thread, i.e. it must mean that `&T` is `Send` and thus it must mean that `T` is `Sync`.

Amanieu: The `ArcMutexGuard` is more like a `&mut T` actually, which also only requires `T: Send` for `&mut T: Send`. This means that if you send the guard, there is still only one thread accessing the `&T` at any one point so `T: !Sync` is still fine.

PR #352, opened 2022-09-02
Document async signal safety of unpark methods

mqudsi:
...without mentioning the word "async" since for majority of rust developers, it
has an entirely different meaning (so does "safety" but I think this is good
enough).

Safety disclaimer is only added to the unpark methods, as those are the only
ones that one might sanely wish to call from an async context.

Happy to tweak the wording if someone has a better suggestion.

### DIFF
--- a/core/src/parking_lot.rs
+++ b/core/src/parking_lot.rs
@@ -724,6 +724,10 @@ pub unsafe fn park(
 ///
 /// The `callback` function is called while the queue is locked and must not
 /// panic or call into any function in `parking_lot`.
+///
+/// The `parking_lot` functions are not re-entrant and calling this method
+/// from the context of an asynchronous signal handler may result in undefined
+/// behavior, including corruption of internal state and/or deadlocks.
 #[inline]
 pub unsafe fn unpark_one(
     key: usize,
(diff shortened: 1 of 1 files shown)


Amanieu:
I think the wording should be stronger about the consequences: since we use thread-local storage, you could get UB instead of just deadlocks.

mqudsi:
How about

> Calling this method from the context of an asynchronous signal handler will result in undefined behavior that may include corruption of internal state and/or deadlocks, unless all signals are masked prior to the {invocation, function call}.

Amanieu:
I think we should just stick to saying that the function isn't re-entrant. There are many ways that users can ensure that the interrupted code is not in the middle of parking_lot, and we shouldn't prefer any specific one.

mqudsi:
So something like this?

> The `parking_lot` functions are not re-entrant and calling this method from the context of an asynchronous signal handler will result in undefined behavior that may include corruption of internal state and/or deadlocks.

Amanieu:
I'd replace "will result" with "may result" but otherwise this looks fine. I'd go with the first version that applies to all `parking_lot` functions.

mqudsi:
Branch updated.

Amanieu:
Thanks!

PR #353, opened 2022-09-04
Update feature advantages over std

yjhn:
`Mutex`, `RwLock` and `CondVar` are const-constructible in `std` since [Rust 1.63](https://blog.rust-lang.org/2022/08/11/Rust-1.63.0.html#const-mutex-rwlock-condvar-initialization). In most platforms `pthread` locks have been replaced with ones that do not allocate.

Maybe it should be noted that points 2-5 are implemented in `std` and 1 is partially implemented?


Amanieu:
I think point 2 can be removed entirely.

Amanieu:
Thanks!

PR #355, opened 2022-09-20
Fix `ReentrantMutexGuard::try_map` signature

### LINKED ISSUE #354: ReentrantMutexGuard::try_map exposes unsound mutable borrow
A ReentrantMutex is supposed to only directly expose immutable references to its contents in order to allow multiple overlapping borrows from the same thread. However, as of lock_api 0.4.8 [`ReentrantMutexGuard::<T>::try_map`](https://docs.rs/lock_api/0.4.8/lock_api/struct.ReentrantMutexGuard.html#method.try_map) exposes a mutable reference to `T`; it calls a function of type `F: FnOnce(&mut T) -> Option<&mut U>`.

It looks to me like this was some kind of porting error that retained the signature from `MutexGuard::try_map`. The mutability isn't called out in the docs and the implementation doesn't account for the case where another remutex guard is borrowing the data, as far as I can see. The rest of the `{Mapped,}ReentrantMutexGuard::{try_,}map` functions all go from `&T` to `&U`, as I would expect.

Amanieu:
This previously incorrectly used `&mut T` instead of `&T`.

Fixes #354

This is a breaking change but will be released in a minor version since it fixes a soundness issue.

### DIFF
--- a/core/src/thread_parker/generic.rs
+++ b/core/src/thread_parker/generic.rs
@@ -8,8 +8,8 @@
 //! A simple spin lock based thread parker. Used on platforms without better
 //! parking facilities available.
 
-use core::sync::atomic::{AtomicBool, Ordering};
 use core::hint::spin_loop;
+use core::sync::atomic::{AtomicBool, Ordering};
 use std::thread;
 use std::time::Instant;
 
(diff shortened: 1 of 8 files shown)


Imberflur:
Would it be reasonable to yank `0.4.8`?

Imberflur:
Ah! The wording in the associated issue had me thinking it was just `0.4.8`. Thanks for clarifying.

PR #356, opened 2022-10-03
Update `windows-sys` to `0.42`

a1phyr:


lazka:
windows-sys 0.42 has some fixes required for us to build for rust+mingw on arm64. It would be nice if this new version would be adopted if possible :)

Amanieu:
Published parking_lot_core 0.9.4.

lazka:
Thanks!

PR #357, opened 2022-10-09
Add pinned variant of methods

notgull:
Unfortunately, there is currently no good API for getting a pinned mutable reference to the inside of a `Mutex` or `RwLock` if you are given a pinned reference to the lock itself. This means that it's not possible to get a `Pin<MutexGuard<T>>` and then a `Pin<&mut T>` when given a `Pin<&Mutex<T>>`. This is noticable, e.g, when implementing something like a shared concurrent `Future`.

```
struct Shared<F>(Arc<Mutex<F>>);

impl<F: Future> Future for Shared<F> {
    type Output = F::Output;

    fn poll(self: Pin<&mut Self>, ...) {
        // Assume that we can safely preform pin projection through pin-project.
        let this = self.project();

        // What I want to do:
        match this.0.try_lock_pinned() { /* ... */ }

        // What I currently do:
        let mtx = unsafe { Pin::into_inner_unchecked(this.0.as_ref()) };
        let lock = mtx.try_lock().map(|lock| unsafe { Pin::new_unchecked(lock) });
        match lock { /* ... */ }
    }
}
```

I need to manually map the contents of the `Mutex` such that I am able to acquire an `Option<Pin<MutexGuard>>`. This requires unsafe code. This PR aims to ameliorate this by providing pinned variants of these methods.

The only issue I see here is with API sprawl, since this PR essentially doubles the current API surface of `lock_api`.

### DIFF
--- a/lock_api/src/mutex.rs
+++ b/lock_api/src/mutex.rs
@@ -10,6 +10,7 @@ use core::fmt;
 use core::marker::PhantomData;
 use core::mem;
 use core::ops::{Deref, DerefMut};
+use core::pin::Pin;
 
 #[cfg(feature = "arc_lock")]
 use alloc::sync::Arc;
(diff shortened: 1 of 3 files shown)


lock_api/src/mutex.rs
```diff
@@ -233,6 +248,20 @@ impl<R: RawMutex, T: ?Sized> Mutex<R, T> {
         }
     }
 
+    /// Attempts to acquire a pinned mutex and returns a pinned guard.
+    /// 
+    /// This function is equivalent to the [`try_lock()`] function, but takes a `Pin<&Mutex>`
+    /// and returns a `Pin<MutexGuard>`. This can be useful if you want to use
+    /// the mutex in a context where the data it protects is pinned, such as a future.
+    #[inline]
+    pub fn try_lock_pinned(self: Pin<&Self>) -> Option<Pin<MutexGuard<'_, R, T>>> {
```
bjorn3, line 257: This is unsound. `&T` is never pinned. As such if you have `Pin<&Mutex<R, TypeThatNeedsPin>>` then you can simply do `pinned_mutex.try_lock()` to get an `Option<MutexGuard<'_, R, TypeThatNeedsPin>>`, which then allows you to move the value that was supposed to be pinned, thus breaking the pin invariant.

notgull: This returns a `Pin<MutexGuard<T>>`, not a `MutexGuard<T>`, which is equivalent to a `Pin<&mut T>` and cannot be moved out of soundly.

notgull: Wait, I see what youre saying. This is unsound. Closing.

PR #358, opened 2022-10-10
Use macros to reduce code deduplication

notgull:
In #291, I added `Arc`-oriented Mutex locks; however, when writing the locking functions for this code, there is a lot of code that ends up being duplicated. This PR adds a macro that reduces the amount of code written (and may make it easier to add more `Arc`-oriented methods in the future.

I'm not particularly attached to any kind of macro syntax.


Amanieu:
I'm not a big fan of this change. I think it makes maintenance harder because the code is now much less readable and it isn't formatted by rustfmt.

notgull:
Fair enough.

PR #362, opened 2022-11-14
Update to windows 0.43

Jasper-Bekkers:
A new windows-rs version came out so we all have to upgrade again.


Jasper-Bekkers:
False alarm, looks like only windows-rs got updated, not windows-sys

PR #363, opened 2022-11-22
Update use of `libc::timespec` to prepare for future libc version

wesleywiser:
In a future release of the `libc` crate, `libc::timespec` will contain private padding fields on 32-bit `*-linux-musl` targets and so the struct will no longer be able to be created using the literal initialization syntax.

Update the use of `libc::timespec` to create a value by calling `std::mem::zeroed()` first and then setting the `tv_sec` and `tv_nsec` fields manually which works with both current versions of `libc` as well as the future version which contains that change.

See also https://github.com/rust-lang/libc/pull/2088

### DIFF
--- a/core/src/thread_parker/linux.rs
+++ b/core/src/thread_parker/linux.rs
@@ -80,10 +80,10 @@ impl super::ThreadParkerT for ThreadParker {
                 self.park();
                 return true;
             }
-            let ts = libc::timespec {
-                tv_sec: diff.as_secs() as libc::time_t,
-                tv_nsec: diff.subsec_nanos() as tv_nsec_t,
-            };
+            // SAFETY: libc::timespec is zero initializable.
+            let mut ts: libc::timespec = std::mem::zeroed();
+            ts.tv_sec = diff.as_secs() as libc::time_t;
+            ts.tv_nsec = diff.subsec_nanos() as tv_nsec_t;
             self.futex_wait(Some(ts));
         }
         true


wesleywiser:
I don't know why this is failing on `macos` but it looks unrelated to this change:

```
  = note: ld: in /Users/runner/.rustup/toolchains/1.49.0-x86_64-apple-darwin/lib/rustlib/x86_64-apple-darwin/lib/libpanic_unwind-f52081aedccc205f.rlib(lib.rmeta), archive member 'lib.rmeta' with length 31176 is not mach-o or llvm bitcode file '/Users/runner/.rustup/toolchains/1.49.0-x86_64-apple-darwin/lib/rustlib/x86_64-apple-darwin/lib/libpanic_unwind-f52081aedccc205f.rlib'
          clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

bjorn3:
All rustc versions before 1.54 are broken with the latest XCode version as the linker no longer accepts non-object files in archives. https://github.com/rust-lang/rust/pull/84449 which landed in 1.54 started wrapping the crate metadata in an object file. See https://github.com/rust-lang/rust/issues/103044 for more info.

Amanieu:
CI should be fixed now, can you rebase?

wesleywiser:
Sure, rebased!

wesleywiser:
Looks like CI is green! 🎉

PR #367, opened 2023-01-04
Add support for watchOS.

pixlwave:
This PR adds support to parking_lot for watchOS. This can be tested by building on a mac with the following command when the watchOS SDK is installed through Xcode:
```
cargo build -Zbuild-std --target=arm64_32-apple-watchos
```
The support simply mirrors macOS and iOS support.

Note: I've also got a `watchos-0.8` branch with the changes cherry picked onto `backport-core-0.8`, I'm not exactly sure how backports are handled.

### DIFF
--- a/core/src/thread_parker/unix.rs
+++ b/core/src/thread_parker/unix.rs
@@ -5,7 +5,7 @@
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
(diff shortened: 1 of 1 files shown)


Amanieu:
Would it make sense to add `tvos` as well?

Regarding backports, I prefer not to unless there is a very good reason. Generally speaking the only supported version of this crate is the latest published version. If you need support for watchOS then you should upgrade your dependencies to the latest version.

PR #369, opened 2023-02-02
Mark `MutexGuard` with `#[clippy::has_significant_drop]`

c410-f3r:
`#[clippy::has_significant_drop]` tells that a structure should be considered when evaluating some lints. Examples of such behavior are the existent `clippy::significant_drop_in_scrutinee` and the soon-to-be-finished https://github.com/rust-lang/rust-clippy/issues/9399.

### DIFF
--- a/lock_api/src/mutex.rs
+++ b/lock_api/src/mutex.rs
@@ -485,7 +485,7 @@ where
 ///
 /// The data protected by the mutex can be accessed through this guard via its
 /// `Deref` and `DerefMut` implementations.
-#[clippy::has_significant_drop]`
+#[clippy::has_significant_drop]
 #[must_use = "if unused the Mutex will immediately unlock"]
 pub struct MutexGuard<'a, R: RawMutex, T: ?Sized> {
     mutex: &'a Mutex<R, T>,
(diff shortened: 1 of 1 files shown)


lock_api/src/mutex.rs
```diff
@@ -485,6 +485,7 @@ where
 ///
 /// The data protected by the mutex can be accessed through this guard via its
 /// `Deref` and `DerefMut` implementations.
+#[clippy::has_significant_drop]`
```
Amanieu, line 488: Stray ` character

c410-f3r: Thanks

mkroening:
@c410-f3r, should we also apply this to all the other guards, like the `RwLock` guards and the reentrant mutex guard, as well as their mapped variants?

c410-f3r:
Yes, any lock primitive should apply `#[clippy::has_significant_drop]` in order to trigger the two mentioned lints.

Moreover, it is very likely that other lints will also use ` #[clippy::has_significant_drop]` in the future.

mkroening:
I opened https://github.com/Amanieu/parking_lot/pull/371.

c410-f3r:
Nice! Thank you

PR #374, opened 2023-03-20
Remove `windows-sys` dependency

Jake-Shadle:
This replaces the `windows-sys` dependency with manual bindings to only the functions, types, and constants needed by `parking_lot`.

## Rationale

`parking_lot` is a foundational dependency, and thus its choice of dependencies has a fairly wide effect on the ecosystem since it's almost inevitable that even a moderately sized dependency graph will pull it in. This is particularly true due the problematic issue I pointed out in https://github.com/microsoft/windows-rs/issues/1720, in which I actually used parking_lot as an example due to its foundational nature. Using bindings in the few places they are needed means we can completely get rid of the `windows-sys` dependency, positively impacting a huge part of the crates ecosystem with no negatives.

1. Win32 functions/types/constants **will never change**, there is no reason to use an external crate of bindings
2. `parking_lot_core` already has manual bindings to `Nt*` functions since they are not officially supported by Microsoft and thus not part of the Win32 metadata used to generate the windows-sys bindings.

### DIFF
--- a/core/src/thread_parker/windows/bindings.rs
+++ b/core/src/thread_parker/windows/bindings.rs
@@ -1,0 +1,32 @@
+pub const INFINITE: u32 = 4294967295;
+pub const ERROR_TIMEOUT: u32 = 1460;
+pub const GENERIC_READ: u32 = 2147483648;
+pub const GENERIC_WRITE: u32 = 1073741824;
+pub const STATUS_SUCCESS: i32 = 0 as _;
+pub const STATUS_TIMEOUT: i32 = 258 as _;
+
+pub type HANDLE = isize;
+pub type HINSTANCE = isize;
+pub type BOOL = i32;
+pub type BOOLEAN = u8;
+pub type NTSTATUS = i32;
+pub type FARPROC = Option<unsafe extern "system" fn() -> isize>;
+pub type WaitOnAddress = unsafe extern "system" fn(
+    Address: *const std::ffi::c_void,
+    CompareAddress: *const std::ffi::c_void,
+    AddressSize: usize,
+    dwMilliseconds: u32,
+) -> BOOL;
+pub type WakeByAddressSingle = unsafe extern "system" fn(Address: *const std::ffi::c_void);
+
+#[link(name = "kernel32")]
+extern "system" {
+    pub fn GetLastError() -> u32;
+    pub fn CloseHandle(hObject: HANDLE) -> BOOL;
+
+    pub fn GetModuleHandleA(lpModuleName: *const u8) -> HINSTANCE;
+    pub fn GetProcAddress(hModule: HINSTANCE, lpProcName: *const u8) -> FARPROC;
+
+    pub fn Sleep(dwMilliseconds: u32);
+}
+
--- a/core/src/thread_parker/windows/mod.rs
+++ b/core/src/thread_parker/windows/mod.rs
@@ -11,6 +11,7 @@ use core::{
 };
 use std::time::Instant;
 
+mod bindings;
 mod keyed_event;
 mod waitaddress;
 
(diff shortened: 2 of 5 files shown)


core/src/thread_parker/windows/mod.rs
```diff
@@ -60,7 +61,7 @@ impl Backend {
             Err(global_backend_ptr) => {
                 unsafe {
                     // We lost the race, free our object and return the global one
-                    Box::from_raw(backend_ptr);
+                    let _ = Box::from_raw(backend_ptr);
                     &*global_backend_ptr
```
Jake-Shadle, line 65: Clippy fix

core/src/thread_parker/windows/mod.rs
```diff
@@ -163,26 +164,10 @@ impl UnparkHandle {
...
-    // libsynchronization.a library pulls in a dependency on a newer DLL not
-    // present in older versions of Windows. (see rust-lang/rust#49438)
-    //
-    // This is a bit of a hack for now and ideally we'd fix MinGW's own import
-    // libraries, but that'll probably take a lot longer than patching this here
-    // and avoiding the `synchapi` feature entirely.
-    extern "system" {
-        fn Sleep(a: u32);
-    }
     unsafe {
```
Jake-Shadle, line 167: This comment was referring to the previous winapi crate usage, and was misleading and irrelevant.

core/src/thread_parker/windows/waitaddress.rs
```diff
@@ -10,30 +10,17 @@ use core::{
...
-    WaitOnAddress: extern "system" fn(
-        Address: *mut ffi::c_void,
-        CompareAddress: *mut ffi::c_void,
-        AddressSize: usize,
-        dwMilliseconds: u32,
-    ) -> BOOL,
-    WakeByAddressSingle: extern "system" fn(Address: *mut ffi::c_void),
+    WaitOnAddress: WaitOnAddress,
+    WakeByAddressSingle: WakeByAddressSingle,
 }
```
Jake-Shadle, line 19: These functions were incorrectly not marked `unsafe`.

core/src/thread_parker/windows/waitaddress.rs
```diff
@@ -10,30 +10,17 @@ use core::{
...
 
 impl WaitAddress {
     #[allow(non_snake_case)]
     pub fn create() -> Option<WaitAddress> {
-        // MSDN claims that that WaitOnAddress and WakeByAddressSingle are
-        // located in kernel32.dll, but they are lying...
         let synch_dll = unsafe { GetModuleHandleA(b"api-ms-win-core-synch-l1-2-0.dll\0".as_ptr()) };
```
Jake-Shadle, line 24: This comment is outdated, MSDN does show this as the library that implements those functions.

core/src/thread_parker/windows/bindings.rs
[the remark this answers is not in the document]
Jake-Shadle: Done!

core/src/thread_parker/windows/bindings.rs
```diff
@@ -0,0 +1,37 @@
...
+//! as these bindings will **never** change and parking_lot_core is a foundational
+//! dependency for the Rust ecosystem, so the dependencies used by it have an
+//! outsize affect
+
+pub const INFINITE: u32 = 4294967295;
+pub const ERROR_TIMEOUT: u32 = 1460;
+pub const GENERIC_READ: u32 = 2147483648;
+pub const GENERIC_WRITE: u32 = 1073741824;
+pub const STATUS_SUCCESS: i32 = 0 as _;
+pub const STATUS_TIMEOUT: i32 = 258 as _;
```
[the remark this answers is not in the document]
Jake-Shadle: Yah, noticed this as well (I'm using a tool I created to generate these bindings), fixed!

PR #396, opened 2023-08-09
windows: impl `ThreadParkerT` / `UnparkHandleT`.

waywardmonkeys:
These imports were introduced in 2fb27d35 and were unused at the time and remain so.

### DIFF
--- a/core/src/thread_parker/windows/mod.rs
+++ b/core/src/thread_parker/windows/mod.rs
@@ -75,11 +75,13 @@ pub struct ThreadParker {
     backend: &'static Backend,
 }
 
-impl ThreadParker {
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
Ah, it was unused on Windows, but not Linux, because `src/thread_parker/windows/mod.rs` has those functions implemented by `ThreadParker` directly rather than `ThreadParkerT` (and same for `UnparkHandleT` stuff).

waywardmonkeys:
The initial version of this patch did something different as I had misunderstood the problem. When building on Windows, there were some warnings about unused imports and I mistakenly throught the imports were unused rather than realizing that the actual issue was that the Windows code wasn't correctly implementing those traits.

PR #403, opened 2023-09-21
Allow unsafe creation of new lock guards

Lochlanna:
This PR changes the visibility of the functions that produce the lock guards to public. 
The inspiration for this comes from my desire to extend the mutex with new locking functions (async locking for example).

I was initially concerned that exposing this might be a bad idea even if it is useful. However, I decided that exposing these functions was no worse than the already public `force_unlock` functions.

### DIFF
--- a/lock_api/src/mutex.rs
+++ b/lock_api/src/mutex.rs
@@ -309,7 +309,7 @@ impl<R: RawMutex, T: ?Sized> Mutex<R, T> {
     /// the guard was forgotten with `mem::forget`.
     #[cfg(feature = "arc_lock")]
     #[inline]
-    unsafe fn make_guard_arc_unchecked(self: &Arc<Self>) -> ArcMutexGuard<R, T> {
+    unsafe fn make_arc_guard_unchecked(self: &Arc<Self>) -> ArcMutexGuard<R, T> {
         ArcMutexGuard {
             mutex: self.clone(),
             marker: PhantomData,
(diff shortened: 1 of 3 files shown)


Amanieu:
Could you rename these functions to `make_guard_unchecked` so that it's clear that these functions are unsafe and that you should carefully read the documentation for them?

lock_api/src/mutex.rs
```diff
@@ -191,9 +191,11 @@ impl<R, T> Mutex<R, T> {
 impl<R: RawMutex, T: ?Sized> Mutex<R, T> {
     /// # Safety
```
Amanieu, line 192: This needs proper documentation, a line that explains what the function does.

lock_api/src/rwlock.rs
```diff
@@ -411,9 +411,11 @@ impl<R, T> RwLock<R, T> {
 impl<R: RawRwLock, T: ?Sized> RwLock<R, T> {
     /// # Safety
     ///
-    /// The lock must be held when calling this method.
+    /// The lock must be held when calling this method. This method must only
+    /// be called if this thread holds the read lock and no `RwLockReadGuard` exists
+    /// for this lock
```
Amanieu, line 416: In the case of read locks, other guards can exist.

Amanieu (commented):
This is missing methods for other guard types such as upgradable guards and Arc guards.

Amanieu:
Slight nit: could you rename the arc versions to `make_arc(_read/write/upgradable)_guard`?

After that it should be good to merge and I will publish a release to address #406

Amanieu:
Thanks!

Issue #418, opened 2023-11-06
Mutex performance drop in case of cache contention


pingzhaozz:
There's a problem of parking-lot Mutex when met cache line contention, the performance dropped much. It can also be observed with [lock-bench](https://github.com/matklad/lock-bench) 

```
$ cargo run --release 32 2 10000 100
Options {
    n_threads: 32,
    n_locks: 2,
    n_ops: 10000,
    n_rounds: 100,
}

std::sync::Mutex     avg 113.030539ms min 105.760154ms max 131.87258ms
parking_lot::Mutex   avg 403.756547ms min 326.026509ms max 533.260014ms
spin::Mutex          avg 161.125953ms min 151.708034ms max 177.132377ms
AmdSpinlock          avg 158.042233ms min 148.723058ms max 171.265994ms

```

It's observed on INTEL(R) XEON(R) PLATINUM 8580 w/ 120 cores

Debug shows currently `spin()` strategy will enter Parking state after "spinwait" failure. The Parking mechanism introduces an overhead of approximately `~100ms`(from lock-bench data). In case of cache contention in multi-core, multi-thread scenarios, the likelihood of `spin()` failures is significantly higher, leading to longer lock durations(there's another lock in parking thread list which may meet cache contention too). Considering the `~100ms` overhead of parking and the millisecond-level or even lower `<1ms` durations of spin, there needs to be a buffer transition between them to avoid the performance loss caused by frequent entries into the parking state.

Currently `yield` is used in spinwait. When there are multiple threads running on the same core, `yield` can effectively alleviate contention problem. However, when the scheduler's ready queue contains only the current thread, the `yield` effect is minimal and may not effective . A possible way is adding some `sleep` before parking and after `spin()`. The result shows good which is better than using `mm_pause` or `yield`. It improves the lock bench much and shows better cpu utility. I'll submit a PR later.

Sleep 1ms before parking:
```
$cargo run --release 32 2 10000 100
Options {
    n_threads: 32,
    n_locks: 2,
    n_ops: 10000,
    n_rounds: 100,
}

std::sync::Mutex     avg 113.276158ms min 103.870893ms max 131.823024ms
parking_lot::Mutex   avg 81.669426ms  min 72.584055ms  max 88.3535ms
spin::Mutex          avg 161.586476ms min 152.302867ms max 184.132674ms
AmdSpinlock          avg 157.446488ms min 147.091038ms max 180.832205ms

```

Issue #338, opened 2022-04-23
Mutex is slower than standard library's  on Intel i9 and ARM CPUs


jerry73204:
I run the mutex benchmark using this command for 36 system cores for example:

```sh
cargo run --bin mutex --release -- 9:36:9 5 5 2 2
```

The parking\_lot's mutex is faster only on Intel CPU and with smaller number of threads. It gets drastically slower when full cores/hyperthreads are utilized. Please tell me if there was anything done wrong.

Intel i7-10750H (6 cores, 12 threads)
```
- Running with 3 threads
- 5 iterations inside lock, 5 iterations outside lock
- 2 seconds per test
        name         |    average     |     median     |    std.dev.
parking_lot::Mutex   |   8521.370 kHz |   8489.291 kHz |    512.243 kHz
std::sync::Mutex     |   4192.444 kHz |   4216.470 kHz |    145.741 kHz
pthread_mutex_t      |   4985.169 kHz |   5060.596 kHz |    233.319 kHz
- Running with 6 threads
        name         |    average     |     median     |    std.dev.
parking_lot::Mutex   |   2094.819 kHz |   2091.441 kHz |    109.118 kHz
std::sync::Mutex     |   1304.931 kHz |   1310.466 kHz |     23.064 kHz
pthread_mutex_t      |   1672.081 kHz |   1687.514 kHz |     61.920 kHz
- Running with 9 threads
        name         |    average     |     median     |    std.dev.
parking_lot::Mutex   |    924.263 kHz |    925.953 kHz |     34.467 kHz
std::sync::Mutex     |    866.749 kHz |    868.601 kHz |     33.733 kHz
pthread_mutex_t      |   1085.478 kHz |   1092.233 kHz |     46.641 kHz
- Running with 12 threads
        name         |    average     |     median     |    std.dev.
parking_lot::Mutex   |    470.583 kHz |    472.499 kHz |      8.968 kHz
std::sync::Mutex     |    638.854 kHz |    640.303 kHz |      7.311 kHz
pthread_mutex_t      |    809.968 kHz |    809.067 kHz |     15.230 kHz
```

Intel i9-10980XE (18 cores, 36 threads)
```
    Finished release [optimized] target(s) in 0.01s
     Running `target/release/mutex '9:36:9' 5 5 2 2`
- Running with 9 threads
- 5 iterations inside lock, 5 iterations outside lock
- 2 seconds per test
        name         |    average     |     median     |    std.dev.
[...]

PR #419, opened 2023-11-06
Fix issue #418.

pingzhaozz:
To avoid entering Parking too frequently in case of cache contention, adding sleep 1ms, 4 times before parking and after old 'spin()'.

The lock-bench result shows:
```
$cargo run --release 32 2 10000 100
Options {
    n_threads: 32,
    n_locks: 2,
    n_ops: 10000,
    n_rounds: 100,
}

std::sync::Mutex     avg 113.276158ms min 103.870893ms max 131.823024ms
parking_lot::Mutex   avg 81.669426ms  min 72.584055ms  max 88.3535ms
spin::Mutex          avg 161.586476ms min 152.302867ms max 184.132674ms
AmdSpinlock          avg 157.446488ms min 147.091038ms max 180.832205ms
```

vs. Before:

```
$ cargo run --release 32 2 10000 100
Options {
    n_threads: 32,
    n_locks: 2,
    n_ops: 10000,
    n_rounds: 100,
}

std::sync::Mutex     avg 113.030539ms min 105.760154ms max 131.87258ms
parking_lot::Mutex   avg 403.756547ms min 326.026509ms max 533.260014ms
spin::Mutex          avg 161.125953ms min 151.708034ms max 177.132377ms
AmdSpinlock          avg 158.042233ms min 148.723058ms max 171.265994ms
```

### DIFF
--- a/core/src/spinwait.rs
+++ b/core/src/spinwait.rs
@@ -6,7 +6,8 @@
 // copied, modified, or distributed except according to those terms.
 
 use crate::thread_parker;
-use core::hint::spin_loop;
+use core::{hint::spin_loop, time::Duration};
+use std::thread;
 
 // Wastes some CPU time for the given number of iterations,
 // using a hint to indicate to the CPU that we are spinning.
@@ -44,16 +45,20 @@ impl SpinWait {
     ///
     /// The spin strategy will initially use a CPU-bound loop but will fall back
     /// to yielding the CPU to the OS after a few iterations.
+    /// Before parking, 'spin()' will finally try sleep 1ms, 4 times to avoid
+    /// entering parking too frequently in case of cache contention.
     #[inline]
     pub fn spin(&mut self) -> bool {
-        if self.counter >= 10 {
+        if self.counter >= 15 {
             return false;
         }
         self.counter += 1;
         if self.counter <= 3 {
             cpu_relax(1 << self.counter);
-        } else {
+        } else if self.counter <= 10 {
             thread_parker::thread_yield();
+        } else {
+            thread::sleep(Duration::from_millis(1));
         }
         true
     }


pingzhaozz:
@Amanieu Sorry to push. I would like to hear some feedbacks about issue #418 and this PR. Will you take a look on it?

Amanieu:
The benchmark numbers here are actually very misleading. Essentially, this benchmark measures how fast a thread can repeatedly acquire and release a lock.

The best performance is obtained when only 1 thread is accessing the lock memory, so there are no cache conflicts, no need to fall back to the slow path, etc. This is the number you get when running the benchmark with only 1 thread.

When running the benchmark with multiple threads, you get the best performance in the same way: by ensuring a single thread can repeatedly lock/unlock without interference from other threads. The easiest way to achieve this is to put all other threads to sleep. The longer the sleep, the better the benchmark numbers.

The problem is that this only makes the benchmark look good on paper, and actually hurts real applications. Making a thread sleep for 1ms (~3000000 CPU cycles) if it misses a lock can have very noticeable impacts on application performance, especially for timing-sensitive code such as rendering or audio. Instead, it is better to park the thread so that it can be woken up by the OS as soon as the lock is free.

pingzhaozz:
The Mutex is designed for multi-threaded data access, ideally program without the need for data sharing. However, in practice, some critical structures require sharing among multiple threads. Based on our experience, this lock benchmark can effectively represent scenarios of lock contention, as illustrated by the given case reflecting Mutex behavior in high contention situations. While reducing the number of threads will have better results, the issue persists, resulting in lower performance compared to std::Mutex. I have chosen this case to better highlight the problem.   

  The performance drop in parking-lot::Mutex, compared to std::Mutex, occurs in the slow path. Before entering the parking state, there should be a smooth transition, and using sleep appears to be a suitable way to transition between spin_wait, yield, and parking. The logic remains consistent with before, meaning that the slow path still attempts spin_wait and yield first. If these also fail and before entering the parking state, it will try sleeping to make the transition smoother. This is because parking takes the OS significantly more time to wake than a simple sleep. Based on that it won’t hurt the timing-sensitive program but help them smoothly transition in the case of high lock contention.

pingzhaozz:
This change also fixe issue #338 

Before:
```
- Running with 9 threads
- 5 iterations inside lock, 5 iterations outside lock
- 2 seconds per test
parking_lot::Mutex   |    363.574 kHz |    330.716 kHz |    137.608 kHz
std::sync::Mutex     |    699.105 kHz |    613.367 kHz |    223.500 kHz
pthread_mutex_t      |   1125.346 kHz |   1115.156 kHz |     46.531 kHz
- Running with 18 threads
parking_lot::Mutex   |     26.346 kHz |     26.358 kHz |      0.407 kHz
std::sync::Mutex     |    353.564 kHz |    336.981 kHz |     61.858 kHz
pthread_mutex_t      |    529.260 kHz |    529.450 kHz |     19.034 kHz
- Running with 27 threads
parking_lot::Mutex   |     17.232 kHz |     17.306 kHz |      0.428 kHz
std::sync::Mutex     |    226.792 kHz |    213.642 kHz |     41.591 kHz
pthread_mutex_t      |    315.459 kHz |    316.935 kHz |     19.549 kHz
- Running with 36 threads
parking_lot::Mutex   |     12.920 kHz |     12.921 kHz |      0.287 kHz
std::sync::Mutex     |    167.901 kHz |    158.800 kHz |     34.085 kHz
pthread_mutex_t      |    221.542 kHz |    212.404 kHz |     17.621 kHz

```
After
```
- Running with 9 threads
- 5 iterations inside lock, 5 iterations outside lock
- 2 seconds per test
parking_lot::Mutex   |   1179.909 kHz |   1075.771 kHz |    345.058 kHz
std::sync::Mutex     |    808.353 kHz |    853.010 kHz |    128.165 kHz
pthread_mutex_t      |   1382.591 kHz |   1388.152 kHz |     22.390 kHz
- Running with 18 threads
parking_lot::Mutex   |    571.435 kHz |    539.379 kHz |    157.537 kHz
std::sync::Mutex     |    384.533 kHz |    397.613 kHz |     86.923 kHz
pthread_mutex_t      |    572.527 kHz |    572.878 kHz |     26.068 kHz
- Running with 27 threads
parking_lot::Mutex   |    368.414 kHz |    345.560 kHz |     85.260 kHz
std::sync::Mutex     |    238.103 kHz |    223.207 kHz |     59.118 kHz
pthread_mutex_t      |    337.518 kHz |    344.361 kHz |     23.691 kHz
- Running with 36 threads
parking_lot::Mutex   |    268.471 kHz |    259.280 kHz |     50.276 kHz
std::sync::Mutex     |    175.217 kHz |    166.470 kHz |     41.695 kHz
pthread_mutex_t      |    237.988 kHz |    225.297 kHz |     25.596 kHz

```

dista:
We have a video/audio streaming application build on tokio(which enable parking_lot by default), when parking_lot is enabled, when we use wrk to bench http output of the streaming application, the application is bottlenecked in cpu, no matter
how many threads(32 threads for example, it should reach 3200% at most) we assign to tokio, the cpu of our application can not exceed 600%.  

after disable parking_lot, we can reach the number we anticipated.

the benchmark for parking_lot is very pool in our server(AMD EPYC 7502)
cargo run --release 32 2 10000 100
```
std::sync::Mutex     avg 24.545373ms  min 19.094168ms  max 27.440505ms 
parking_lot::Mutex   avg 437.688227ms min 403.22461ms  max 493.899443ms
spin::Mutex          avg 32.515297ms  min 22.190278ms  max 56.560499ms 
AmdSpinlock          avg 32.379324ms  min 23.217082ms  max 48.166241ms 
```

Amanieu:
@dista Can you open a separate issue for this? Also, would you be able to test some modifications of parking_lot on your application to see how it affects CPU usage?

dista:
> @dista Can you open a separate issue for this? Also, would you be able to test some modifications of parking_lot on your application to see how it affects CPU usage?

issue filed https://github.com/Amanieu/parking_lot/issues/437

PR #420, opened 2023-11-17
Update `windows-targets` 0.52

kennykerr:
This just brings expanded lib support to include coverage for the first update to the `windows-sys` crate in 8 months. https://github.com/microsoft/windows-rs/releases/tag/0.52.0

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -27,7 +27,7 @@ libc = "0.2.95"
 redox_syscall = "0.4"
 
 [target.'cfg(windows)'.dependencies]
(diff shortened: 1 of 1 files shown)


kennykerr:
Ah, looks like you have a lower MSRV. Much like `tokio`, the `windows-targets` crate now expects 1.56 to avoid a number of [workspace compatibility](https://github.com/microsoft/windows-rs/pull/2613) issues. 

I'll just go ahead and close this PR and you can revisit when you're ready to bump your MSRV. 👋

PR #425, opened 2024-01-22
Update dependencies & bump MSRV to 1.56.0

westy92:


src/raw_rwlock.rs
```diff
@@ -346,8 +346,8 @@ unsafe impl lock_api::RawRwLockUpgrade for RawRwLock {
     unsafe fn unlock_upgradable(&self) {
         self.deadlock_release();
         let state = self.state.load(Ordering::Relaxed);
-        if state & PARKED_BIT == 0 {
```
Amanieu, line 349: Please don't do this, it makes the code less readable.

westy92: It was a clippy warning to keep it as-is.
IMO it's not less readable.

Amanieu: I understand that this comes from a clippy lint, but I not particularly fond of that lint. The `if` statements are intentionally nested this way so that the structure of the code is clearer. If you must, add an `#[allow]` to disable this clippy lint.

westy92: Updated to match your preferences. 👍

Amanieu:
Thanks! Can you squash the commits?

westy92:
You should be able to squash when you merge. If you can't, then I'm happy to squash and force push.

westy92:
Thanks much! Looking forward to the next release. 🚀

PR #427, opened 2024-02-21
Fix inaccurate documentation of `wait_while`, `wait_while_until` & `wait_while_for`

Jupeyy:
The current documentation can be interpreted incorrectly. For example if you wait for a condition that is already false, the current documentation suggests that the current thread will still block.

see https://github.com/rust-lang/rust/pull/121405

### DIFF
--- a/src/condvar.rs
+++ b/src/condvar.rs
@@ -407,11 +407,12 @@ impl Condvar {
 
         result
     }
-    /// Blocks the current thread until this condition variable receives a
-    /// notification. If the provided condition evaluates to `false`, then the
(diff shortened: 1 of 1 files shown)


Amanieu:
I think the new docs are actually harder to read than the old ones. Maybe it might be worth copying the latest docs from the standard library's `Condvar` if they are clearer: https://doc.rust-lang.org/nightly/std/sync/struct.Condvar.html#method.wait

Jupeyy:
They are wrong too, and I seem not be able to write docs that are clear enough. Maybe a native speaker or someone with more experience is better suited.
Thanks.

PR #432, opened 2024-03-19
Update redox_syscall dependency version

alex:


alex:
Thanks for the quick merge. Is there any current thought as to when the next release might be?

bjorn3:
Redox OS is in the process of moving from the redox_syscall crate to the libredox crate to provide a more stable ABI while still allowing the syscall ABI to evolve: https://crates.io/crates/libredox

Amanieu:
I don't have a plan to make a release any time soon unless you specifically need it. However it would be nice to move to libredox first if it's ready for use.

bjorn3:
> if it's ready for use.

cc @4lDO2

alex:
Just wanted to pass along a thank you for the release!

PR #433, opened 2024-04-06
Add Apple visionOS support

QuentinPerez:
## Motivation

Add the support of Apple VisionOS  [121419](https://github.com/rust-lang/rust/pull/121419)

## Solution
Add target_os = "visionos"

### DIFF
--- a/core/src/thread_parker/unix.rs
+++ b/core/src/thread_parker/unix.rs
@@ -5,12 +5,7 @@
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
-#[cfg(any(
-    target_os = "macos",
-    target_os = "tvos",
-    target_os = "ios",
-    target_os = "watchos"
-))]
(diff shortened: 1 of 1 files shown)


Amanieu:
Can we just use `target_vendor = "apple"` so we don't have to change this every time Apple releases a new device? It might be necessary

QuentinPerez:
Good Idea !

core/src/thread_parker/unix.rs
```diff
@@ -132,26 +127,12 @@ impl super::ThreadParkerT for ThreadParker {
...
     /// Initializes the condvar to use CLOCK_MONOTONIC instead of CLOCK_REALTIME.
-    #[cfg(any(
-        target_os = "macos",
-        target_os = "ios",
-        target_os = "tvos",
-        target_os = "watchos",
-        target_os = "android",
-        target_os = "espidf"
-    ))]
+    #[cfg(target_vendor = "apple")]
```
Amanieu, line 130: You removed android and espidf here.

PR #435, opened 2024-04-12
stdarch_wasm_atomic_wait instead of stdsimd

### LINKED ISSUE #434: unknown feature `stdsimd`
When building on wasm32-unknown-unknown with the `nightly` feature, parking_lot_core won't compile, with an "unknown feature `stdsimd`" error.

`stdsimd` has been removed. To the best of my understanding, the required WASM atomic operations are behind the `stdarch_wasm_atomic_wait` flag now.

sorcio:
Should fix #434 

I don't know the context of the change so I might be missing something important. I only know that I tested on nightly and it works.

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -49,7 +49,7 @@
         target_family = "wasm",
         target_feature = "atomics"
     ),
-    feature(stdsimd)
+    feature(stdarch_wasm_atomic_wait)
 )]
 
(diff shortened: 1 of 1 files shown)


Brooooooklyn:
Could you release a new version containing this?

Amanieu:
Done.

Brooooooklyn:
thanks a lot

PR #445, opened 2024-06-30
Remove warnings due to new lint on unknown cfgs

tea:
Current nightly build shows several warnings such as:

```
warning: unexpected `cfg` condition name: `tsan_enabled`
   --> core/src/word_lock.rs:293:13
    |
293 |     if cfg!(tsan_enabled) {
    |             ^^^^^^^^^^^^
    |
    = help: consider using a Cargo feature instead
    = help: or consider adding in `Cargo.toml` the `check-cfg` lint config for the lint:
             [lints.rust]
             unexpected_cfgs = { level = "warn", check-cfg = ['cfg(tsan_enabled)'] }
    = help: or consider adding `println!("cargo::rustc-check-cfg=cfg(tsan_enabled)");` to the top of the `build.rs`
    = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
    = note: `#[warn(unexpected_cfgs)]` on by default
```

https://blog.rust-lang.org/2024/05/06/check-cfg.html describes this new lint. I found no way to fix this without probing for Rust version, so probing it is.

Also fixed two other warnings shown by the nightly build.


faern:
Polling for the rustc version is not ideal. Also pulling in a dependency just for this is very sad. Can't we just always declare the custom cfg directly in `Cargo.toml`?

tea:
> Polling for the rustc version is not ideal. Also pulling in a dependency just for this is very sad. Can't we just always declare the custom cfg directly in `Cargo.toml`?

If you mean via `[lints.rust]` then not really. Older versions would warn about unknown section, thus defeating the purpose - it doesn't make that much sense to exchange one warning on newer versions for another on older versions. This was best I could figure out to make it warning-free on the entire versions range. Once mininum Rust version goes up enough, then this could be made into config.

 As for dependency, it is already kind of there; it is new for core but already used for lock_api.

Amanieu:
I don't care about warnings showing up on old Rust versions, it's not a breaking change since the code still compiles. Since 1.79 is now stable, we should just unconditionally emit the `cargo:rustc-check-cfg`.

tea:
> I don't care about warnings showing up on old Rust versions, it's not a breaking change since the code still compiles. Since 1.79 is now stable, we should just unconditionally emit the `cargo:rustc-check-cfg`.

Done. Builds cleanly on 1.56 and current stable and nightly. There are warnings now on 1.64 and probably up to 1.76, like the article says.

PR #454, opened 2024-12-26
Initial (optional) `triomphe::Arc` support

virtualritz:
See subject.

This adds support for swapping out `std::sync::Arc` (`parking_lot`) resp. `alloc::sync::Arc` (`lock_api`) for [`triomphe`](https://docs.rs/triomphe/latest/triomphe/)`::`[`Arc`](https://docs.rs/triomphe/latest/triomphe/struct.Arc.html).

Both crates had a `triomphe` feature added but alternatively I could put this behind a `triomphe_arc_lock` feature instead.
Or even `arc_de_triomphe_lock`, to fully express the presumed naming intent of that crate. :wink:

I need this for a project I'm working on so I thought I may as well open a PR, in case this is useful. If not, just close the PR and ignore.

### DIFF
--- a/src/condvar.rs
+++ b/src/condvar.rs
@@ -526,12 +526,15 @@ impl fmt::Debug for Condvar {
 mod tests {
     use crate::{Condvar, Mutex, MutexGuard};
     use std::sync::mpsc::channel;
+    #[cfg(not(feature = "triomphe"))]
     use std::sync::Arc;
     use std::thread;
     use std::thread::sleep;
     use std::thread::JoinHandle;
     use std::time::Duration;
     use std::time::Instant;
+    #[cfg(feature = "triomphe")]
+    use triomphe::Arc;
 
     #[test]
     fn smoke() {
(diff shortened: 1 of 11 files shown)


src/condvar.rs
```diff
@@ -526,12 +526,15 @@ impl fmt::Debug for Condvar {
 mod tests {
     use crate::{Condvar, Mutex, MutexGuard};
     use std::sync::mpsc::channel;
+    #[cfg(not(feature = "triomphe"))]
```
Amanieu, line 529: There's no need to do this for tests, we can always use `std::sync::Arc` there.

Amanieu (changes requested):
Sorry for the late response, this slipped off my review queue.

Overall, I'm very much in favor of adding `triomphe` support, but the issue with the way the code is written is that it doesn't allow both `std::sync::Arc` and `triomphe::Arc` to be supported at the same time. Instead, I would prefer if methods like `arc_lock` were generic over a trait which is implemented for both `std::sync::Arc` and `triomphe::Arc` (and possibly other Arc variants from `triomphe`).

PR #456, opened 2025-01-10
Add `#[track_caller]` for better panic messages for fallible locks using `lock_api`.

SFBdragon:
This PR marks some of the `lock` functions in `lock_api` with `#[track_caller]` where they might reasonably panic for fallible locks.

I would like to write a `RawMutex` implementation that behaves like a mutable-borrow-only `RefCell`. This is fairly trivial, but the main difference is that `RefCell` leverages `#[track_caller]` and `std::panic::Location::caller` to provide great panic messages when `debug_assertions` are enabled, indicating where the prior borrow occurred when panicking. I'd like to replicate that, but the issue is that `lock_api::Mutex`'s `lock` function doesn't have the `#[track_caller]` attribute which results in error messages saying "could not borrow when previously borrows here: [in the lock_api::Mutex::lock function]".

In general, if a RawMutex's `lock` implementation ever fails AND the writer of the `RawMutex` impl marked the `lock` function with `#[track_caller]`, I think it's much more helpful for any panic to indicate where the user of the `lock_api::Mutex` called `lock` rather than where `lock_api::Mutex` called `lock_api::RawMutex::lock`. The latter is useless information, the former is very useful for... well, making an unsynchronized RefCell-like lock (the RawMutex is !Sync and the Guard is !Send).

### DIFF
--- a/lock_api/src/mutex.rs
+++ b/lock_api/src/mutex.rs
@@ -234,6 +234,7 @@ impl<R: RawMutex, T: ?Sized> Mutex<R, T> {
     ///
     /// This function does not block.
     #[inline]
+    #[track_caller]
     pub fn try_lock(&self) -> Option<MutexGuard<'_, R, T>> {
         if self.raw.try_lock() {
             // SAFETY: The lock is held, as required.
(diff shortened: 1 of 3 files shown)


Amanieu:
Looks good to me! Could you also add these to `RwLock`?

Amanieu:
Thanks!

PR #457, opened 2025-01-13
lock_api: Add MappedArcMutexGuard to mirror MappedMutexGuard

dflemstr:
Add a new type, `MappedArcMutexGuard`, that mirrors the API of `MappedMutexGuard` but with an owned `Arc<>` reference to the underlying mutex while locked.

I didn't find any tests covering `MappedMutexGuard`, so I didn't change the status quo. Instead, I copied the exact same implementation as for `MappedMutexGuard` but with lifetimes removed and `&'a R` references to the raw mutex replaced with `Arc<R>`. I believe this should make the new type at least as sound as the previous `MappedMutexGuard`.

The name was chosen based on the existing pattern of `MappedFairMutexGuard` to arrive at `MappedArcMutexGuard` (as opposed to e.g. `ArcMappedMutexGuard` or whatever).


dflemstr:
I'm going to iterate a bit on the implementation of this PR and test it in various tricky production cases to make sure it is sound, so moving into Draft state

dflemstr:
Gonna do some more testing but I think this PR is ready to get new eyes on it.

PR #461, opened 2025-05-07
fast unlock in contention

BusyJay:
During contention, almost all threads are active on CPU, unlock them fast can make those threads make progress more quickly. This help improve global throughput in high contention a lot.

One shortcoming is that fair unlock is now required be invoked explicitly.

This is an improvement to #418.

PR #462, opened 2025-05-08
even fast unlock in contention

BusyJay:
This is an alternative more aggressive implementation of idea #461.

Compared to #461, this PR
- maintains parked bit on waiter side, so that waker doesn't have to atomic operation twice.
- reset all lock states back to 0 when unlock. This makes fast lock more likely succeed during high contention.
- set PARKED_BIT even waiter is prevented from sleep, so that more threads can be woken up during contention to compete for progress.

Running `cargo run --bin mutex --release -- 9:36:9 5 5 2 2`:

Running with 9 threads

--- | --- | --- | ---
parking_lot::Mutex (this pr)   |    472.082 kHz |    475.676 kHz |     13.820 kHz
parking_lot::Mutex (pr 461)   |   405.134 kHz |    406.469 kHz |      9.105 kHz
parking_lot::Mutex (master)   |    364.841 kHz |    365.127 kHz |     11.269 kHz
std::sync::Mutex     |    769.754 kHz |    767.714 kHz |     23.908 kHz
pthread_mutex_t      |    982.966 kHz |    989.991 kHz |     31.816 kHz

Running with 18 threads

--- | --- | --- | ---
parking_lot::Mutex (this pr)  |    352.643 kHz |    353.062 kHz |      7.842 kHz
parking_lot::Mutex  (pr 461)  |    268.530 kHz |    268.355 kHz |      5.586 kHz
parking_lot::Mutex (master)   |     82.786 kHz |     82.975 kHz |      2.435 kHz
std::sync::Mutex     |    389.199 kHz |    394.549 kHz |     21.358 kHz
pthread_mutex_t      |    482.005 kHz |    489.225 kHz |     26.404 kHz

Running with 27 threads

--|--|--|--
parking_lot::Mutex (this pr)  |    231.691 kHz |    231.891 kHz |      3.014 kHz
parking_lot::Mutex (pr 461)   |    185.802 kHz |    186.233 kHz |      2.598 kHz
parking_lot::Mutex (master)   |     28.246 kHz |     28.306 kHz |      0.443 kHz
std::sync::Mutex     |    280.553 kHz |    280.014 kHz |     10.115 kHz
pthread_mutex_t      |    311.815 kHz |    311.582 kHz |      8.409 kHz

Running with 36 threads

--|--|--|--
parking_lot::Mutex (this pr)  |    157.511 kHz |    157.183 kHz |      1.764 kHz
parking_lot::Mutex (pr 461)   |    134.010 kHz |    133.784 kHz |      1.509 kHz
parking_lot::Mutex (master)   |     22.055 kHz |     22.059 kHz |      0.150 kHz
std::sync::Mutex     |    193.672 kHz |    195.078 kHz |     10.369 kHz
pthread_mutex_t      |    224.436 kHz |    225.767 kHz |     12.334 kHz

Running `cargo run --bin rwlock --release -- 36 9 5 5 2 2`

name | write | read
--|--|--
parking_lot::RwLock (this pr) | 6805.309 kHz | 1334.018 kHz
parking_lot::RwLock (pr 461)  | 6121.347 kHz | 968.373 kHz
parking_lot::RwLock (master)  | 628.062 kHz | 954.938 kHz
seqlock::SeqLock | 648.979 kHz | 152225.000 kHz
pthread_rwlock_t | 1678.253 kHz | 376.558 kHz

Using lock-bench, `cargo run --release 32 2 10000 100`:

name | avg | min | max
--|--|--|--
std::sync::Mutex | 30.795793ms  | 28.369313ms  | 33.668656ms
parking_lot::Mutex (this pr) | 32.460918ms  | 29.497656ms | 34.424228ms
parking_lot::Mutex (pr 461) | 40.800542ms | 37.16543ms | 44.621677ms
parking_lot::Mutex (master) | 206.836045ms | 183.902676ms | 213.697023ms
spin::Mutex | 63.898884ms | 58.45244ms | 74.323676ms
AmdSpinlock | 70.131547ms | 65.356139ms | 83.456119ms

name | avg | min | max
--|--|--|--
std::sync::Mutex | 30.52266ms | 28.69828ms | 34.945486ms
parking_lot::Mutex (this pr) | 31.257888ms | 29.648337ms | 33.955286ms
parking_lot::Mutex (pr 461) | 41.146074ms | 38.453175ms | 42.433051ms
parking_lot::Mutex (master) | 210.387478ms | 187.38791ms | 215.752182ms
spin::Mutex | 62.823716ms | 54.801191ms | 74.31628ms
AmdSpinlock | 68.937325ms | 55.406785ms | 80.83359ms

### DIFF
--- a/src/condvar.rs
+++ b/src/condvar.rs
@@ -6,8 +6,8 @@
 // copied, modified, or distributed except according to those terms.
 
 use crate::mutex::MutexGuard;
-use crate::raw_mutex::{RawMutex, TOKEN_HANDOFF, TOKEN_NORMAL};
-use crate::{deadlock, util};
+use crate::raw_mutex::{RawMutex, TOKEN_HANDOFF, TOKEN_NORMAL, TOKEN_RESTORE_PARKED_BIT};
+use crate::util;
 use core::{
     fmt, ptr,
     sync::atomic::{AtomicPtr, Ordering},
@@ -229,9 +229,10 @@ impl Condvar {
             // If we requeued threads to the mutex, mark it as having
             // parked threads. The RequeueAll case is already handled above.
             if op == RequeueOp::UnparkOneRequeueRest && result.requeued_threads != 0 {
-                unsafe { (*mutex).mark_parked() };
+                TOKEN_RESTORE_PARKED_BIT
+            } else {
+                TOKEN_NORMAL
             }
-            TOKEN_NORMAL
         };
         let res = unsafe { parking_lot_core::unpark_requeue(from, to, validate, callback) };
 
@@ -350,10 +351,10 @@ impl Condvar {
         }
 
         // ... and re-lock it once we are done sleeping
-        if result == ParkResult::Unparked(TOKEN_HANDOFF) {
-            unsafe { deadlock::acquire_resource(mutex as *const _ as usize) };
-        } else {
-            mutex.lock();
+        match result {
+            ParkResult::Unparked(TOKEN_HANDOFF) => unreachable!("can't be handed off"),
+            ParkResult::Unparked(TOKEN_RESTORE_PARKED_BIT) => mutex.lock_contention(),
+            _ => mutex.lock(),
         }
 
         WaitTimeoutResult(!(result.is_unparked() || requeued))
--- a/src/raw_mutex.rs
+++ b/src/raw_mutex.rs
@@ -22,6 +22,9 @@ pub(crate) const TOKEN_NORMAL: UnparkToken = UnparkToken(0);
 // thread directly without unlocking it.
 pub(crate) const TOKEN_HANDOFF: UnparkToken = UnparkToken(1);
 
+// UnparkToken used to indicate that the waiter should restore PARKED_BIT.
+pub(crate) const TOKEN_RESTORE_PARKED_BIT: UnparkToken = UnparkToken(2);
+
 /// This bit is set in the `state` of a `RawMutex` when that mutex is locked by some thread.
 const LOCKED_BIT: u8 = 0b01;
 /// This bit is set in the `state` of a `RawMutex` just before parking a thread. A thread is being
@@ -69,7 +72,7 @@ unsafe impl lock_api::RawMutex for RawMutex {
             .compare_exchange_weak(0, LOCKED_BIT, Ordering::Acquire, Ordering::Relaxed)
             .is_err()
         {
-            self.lock_slow(None);
+            self.lock_slow(None, false);
         }
         unsafe { deadlock::acquire_resource(self as *const _ as usize) };
     }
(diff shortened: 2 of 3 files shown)


src/condvar.rs
```diff
@@ -350,10 +351,10 @@ impl Condvar {
         }
 
         // ... and re-lock it once we are done sleeping
-        if result == ParkResult::Unparked(TOKEN_HANDOFF) {
-            unsafe { deadlock::acquire_resource(mutex as *const _ as usize) };
-        } else {
-            mutex.lock();
+        match result {
+            ParkResult::Unparked(TOKEN_HANDOFF) => unreachable!("can't be handed off"),
```
Amanieu, line 355: `TOKEN_HANDOFF` is actually reachable if we are requeued onto a mutex and then another unlocks that mutex with `unlock_fair`.

src/raw_mutex.rs
```diff
@@ -22,6 +22,9 @@ pub(crate) const TOKEN_NORMAL: UnparkToken = UnparkToken(0);
 // thread directly without unlocking it.
 pub(crate) const TOKEN_HANDOFF: UnparkToken = UnparkToken(1);
 
+// UnparkToken used to indicate that the waiter should restore PARKED_BIT.
```
Amanieu, line 25:
```suggestion
// UnparkToken used to indicate that the waiter should restore PARKED_BIT and then continue attempting to acquire the mutex.
```

src/raw_mutex.rs
```diff
@@ -271,12 +276,16 @@ impl RawMutex {
...
 
                 // We were unparked normally, try acquiring the lock again
                 ParkResult::Unparked(_) => (),
 
                 // The validation function failed, try locking again
-                ParkResult::Invalid => (),
+                // This thread doesn't sleep, so it's not sure whether it's the last thread
+                // in queue. Setting PARKED_BIT can lead to false wake up. But false wake up
+                // is good for throughput during high contention.
+                ParkResult::Invalid => extra_flags = PARKED_BIT,
```
Amanieu, line 288: I don't think this is correct: we should only set `PARKED_BIT` if we are required to by an `UnparkToken` or if the current thread is about to park. Otherwise this just causes unnecessary work.

What numbers do you get on the benchmark without this?

Amanieu (commented):
Thanks for this work, I really appreciate it!

I've only reviewed the `RawMutex` part so far, but it looks very promising.

src/raw_mutex.rs
BusyJay: > This has the effect of causing threads to go directly to parking, which as you said is quite slow

Perf stats shows that setting the PARKED_BIT here can lead to more context switch and a lot higher cache miss, this is the prof that more threads are staying on CPU instead of going to sleep.

The reason why PARKED_BIT will wake more threads is because some thread will acquire lock without any competing during contention. For example, if thread a acquire lock, and thread b and c are waiting for a. When a release lock and wake thread b, another thread d that is on CPU right now may acquire lock earlier than thread b. There are two possible behavior of thread d, it can acquire lock directly, or it fails to try lock and try to park but fail again due to validation. Setting parked bit here is utilizing the second situation, so that when d acquire lock, it can still wake thread d later.

> it actually isn't good since other threads are wasting time doing work that isn't useful instead of attempting to acquire the lock.

I notice this performance pitfall when I try to implement a linked list with mutex. After this PR, there is no obvious performance difference between pthread and parking_lot.

Amanieu: In that situation, it's not thread D's job to set the parked bit: thread B will set it before parking itself.

BusyJay: thread B may not as it may be still spinning to try lock.

PR #466, opened 2025-05-25
Fix clearing `PARKED_WRITER_BIT` after a timeout

### LINKED ISSUE #465: `RwLock::try_write_for` corrupts the lock's state to `0xffffffffffffffff` sometimes
We have been investigating a weird lockup in our production system that uses `RwLock` from `parking_lot` this week. After ruling out a bunch of other possible causes, such as a real deadlock, or some kind of UB-induced memory corruption (like a double free), we eventually isolated this down to `RwLock::try_write_for()`.

Here's a test program (courtesy of my colleague @pborzenkov) that seems reproduce this issue somewhat reliably at least on our laptops and servers: https://gist.github.com/PeterCxy/b1307059fc0c06115e08f59f9664f407. There're both an async version and a sync version: the one with Tokio runtime seems to reproduce more reliably for whatever reason. Both contain a feature flag `write_timeout` to enable the use of `try_write_for` -- without that, the bug does not trigger either. Note that both contain some commented-out type punning magic to inspect the internal state  of the lock, but those aren't required to trigger the bug.

Specifically, we were able to track down the issue to this `fetch_add`: https://github.com/Amanieu/parking_lot/blob/master/src/raw_rwlock.rs#L1040. This is done by the following change on top of the current master branch:

```
diff --git a/src/raw_rwlock.rs b/src/raw_rwlock.rs
index 599889b..09dea1f 100644
--- a/src/raw_rwlock.rs
+++ b/src/raw_rwlock.rs
@@ -1041,6 +1041,9 @@ impl RawRwLock {
                         prev_value.wrapping_sub(WRITER_BIT | WRITER_PARKED_BIT),
                         Ordering::Relaxed,
                     );
+                    if state.wrapping_add(prev_value.wrapping_sub(WRITER_BIT | WRITER_PARKED_BIT)) == 0xffffffffffffffff {
+                        panic!("wtf? {state:x}");
+                    }
                     if state & PARKED_BIT != 0 {
                         let callback = |_, result: UnparkResult| {
                             // Clear the parked bit if there no more parked threads
```

...and running the reproducing example against it. I get the following pretty quickly (the async version, just because it's somehow quicker to trigger):

```
thread 'tokio-runtime-worker' panicked at /home/peter/workspace/parking_lot/src/raw_rwlock.rs:1045:25:
wtf? 9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

so the `fetch_add` reported the prior state as `0x9`, which is `0b1001`, corresponding to `WRITER_BIT` and `PARKED_BIT` being set, but most notably, a missing `WRITER_PARKED_BIT`. The code here tries to unset `WRITER_BIT` and `WRITER_PARKED_BIT` by adding `prev_value.wrapping_sub(WRITER_BIT | WRITER_PARKED_BIT)`, but this results in `0xffffffffffffffff` if `WRITER_PARKED_BIT` is not set and `prev_value` is 0.

This is as far as we have got at this point, and we are not sure why `WRITER_PARKED_BIT` would not be set when a thread reaches the `ParkResult::TimedOut` match arm here. Maybe it is somehow possible that there are two threads in progress inside the `RawRwLock::wait_for_readers` function, and they raced in setting/unsetting this bit?

---

Edit: After thinking a bit more about it I wonder whether the story is something like this:

1. Thread 1 grabs the read lock
2. Thread 2 tries to grab the write lock with a timeout, gets parked waiting for reader, sets `WRITER_BIT` and `WRITER_PARKED_BIT`
3. Thread 1 releases read lock, unparks one waiting writer, and thinks it succeeds; it then [unsets `WRITER_PARKED_BIT`](https://github.com/Amanieu/parking_lot/blob/master/src/raw_rwlock.rs#L735)
4. Thread 2 gets woken up, but instead of thinking it unparked successfully, it thinks it timed out; it then tries to unset both `WRITER_BIT` and `WRITER_PARKED_BIT` using the fetch add mentioned in the issue, causing the state to become all 1's.

The problem is I am not sure step 3 and 4 can happen -- can there be a discrepancy between the result of `unpark_one` and what the original thread gets woken with?

Amanieu:
This needs to be done while holding the queue lock to avoid missed wakeups when a concurrent unlock clears this bit.

Fixes #465

### DIFF
--- a/src/raw_rwlock.rs
+++ b/src/raw_rwlock.rs
@@ -1008,7 +1008,11 @@ impl RawRwLock {
                 state & READERS_MASK != 0 && state & WRITER_PARKED_BIT != 0
             };
             let before_sleep = || {};
-            let timed_out = |_, _| {};
+            let timed_out = |_, was_last_thread: bool| {
+                // Clear the parked bit while holding the queue lock. There
+                debug_assert!(was_last_thread);
+                self.state.fetch_and(!WRITER_PARKED_BIT, Ordering::Relaxed);
+            };
             // SAFETY:
             //   * `addr` is an address we control.
             //   * `validate`/`timed_out` does not panic or call into any function of `parking_lot`.
@@ -1037,10 +1041,9 @@ impl RawRwLock {
                     // We need to release WRITER_BIT and revert back to
                     // our previous value. We also wake up any threads that
                     // might be waiting on WRITER_BIT.
-                    let state = self.state.fetch_add(
-                        prev_value.wrapping_sub(WRITER_BIT | WRITER_PARKED_BIT),
-                        Ordering::Relaxed,
-                    );
+                    let state = self
+                        .state
+                        .fetch_add(prev_value.wrapping_sub(WRITER_BIT), Ordering::Relaxed);
                     if state & PARKED_BIT != 0 {
                         let callback = |_, result: UnparkResult| {
                             // Clear the parked bit if there no more parked threads


Amanieu:
@PeterCxy can you check if this fixes your issue?

PeterCxy:
That was quick! I'll test this out in production on Monday and report back.

Thank you!

PeterCxy:
No further lockups in the past >12 hours. For context, this used to happen almost every hour after we started using `try_write_for` before the fix.

PR #486, opened 2025-08-05
Benchmark improvments

### LINKED ISSUE #398: Update README claims
The README makes a number of performance claims which haven't been updated in 7 years, in the meantime the standard library has seen a number of improvements made to core concurrency primitives.  It would be nice to provide a more up-to-date description of why one may or may not want to use `parking_lot`.

unlimitedsola:
I'm trying to resolve  #398 as I strongly agree with [this comment](https://github.com/Amanieu/parking_lot/issues/398#issuecomment-2349247853).

In the process of coming up with benchmark numbers I noticed a few shortcomings of the benchmark crate, notably:

- [The `PthreadRwLock` implementation was calling the wrong function for acquiring the read lock](https://github.com/Amanieu/parking_lot/commit/b1adaefb273d0c8a2e144abf67c05708777dda88)
- [The lack of `RwLock` from stdlib for comparison](https://github.com/Amanieu/parking_lot/commit/79e4a90626a1682b039f3b65490d2f1f8b854160)

The rest of changes are some minor polishments and linting fixes.

---

I'm also planning to add an opinionated benchmark suite runner for making the benchmark numbers more accessible, reproducible, and provable. 

I noticed from your comment that [you're interested in how people benchmark this crate](https://github.com/Amanieu/parking_lot/pull/419#issuecomment-1820171480). Since I'm not an expert in lock implementations, I wanted to use this PR as a chance to ask for your input: what scenarios do you consider important and meaningful for benchmarking, and which ones do you think should be included in the suite to fairly represent this crate?

### DIFF
--- a/benchmark/Cargo.toml
+++ b/benchmark/Cargo.toml
@@ -9,14 +9,6 @@ parking_lot = {path = ".."}
 seqlock = "0.2"
 libc = "0.2"
 
-[[bin]]
-name = "mutex"
-path = "src/mutex.rs"
-
-[[bin]]
-name = "rwlock"
-path = "src/rwlock.rs"
-
 [features]
 nightly = ["parking_lot/nightly"]
 deadlock_detection = ["parking_lot/deadlock_detection"]
--- a/benchmark/src/args.rs
+++ b/benchmark/src/args.rs
@@ -46,12 +46,9 @@ fn print_usage(names: &[&str], error_msg: Option<String>) -> ! {
 }
 
 fn parse_num(names: &[&str], name: &str, value: &str) -> usize {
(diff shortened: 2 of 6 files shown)


unlimitedsola:
> The number one most important benchmark is the performance of uncontended lock acquisition since this is going to be the most common case in practice. This is easy to measure reliably since only one thread is needed.
> 
> Benchmarking situations with contention between multiple threads is more difficult because it is very easy to "cheat" by simply sleeping more between each attempt to acquire the lock when contention is detected. Eventually this devolves into having one thread repeatedly acquiring and releasing the lock while all other threads are just sleeping.

Thank you for the feedback! I'll try to put something together.

PR #243, opened 2020-06-10
lock_api: Mark unlock and friends unsafe.

brain0:
Methods like `unlock`, `bump`, `upgrade`, `downgrade` and similar all assume that a certain kind of lock is being held. Since the wrapper mutex and guard types in the `lock_api` crate guarantee that these APIs are always used correctly, these methods are made unsafe to allow implementers to skip checks for these conditions. This allows sound and efficient implementations of these traits.

I hope I did not forget any methods or screw up any comments.

PR #508, opened 2025-12-28
lock_api: Make `raw` methods safe

purplesyringa:
Since `unlock*` methods are unsafe since #243, it doesn't make much sense to keep `raw` unsafe.

P.S. While removing `unsafe` from uses in `parking_lot`, I've noticed that `parking_lot` seems to play fast-and-loose with whether unlocking is considered `unsafe`: e.g. the `unlock` trait methods are `unsafe`, but the various `unlock_slow`privater methods are safe, and `CondVar::wait_until_internal` is not marked `unsafe` despite taking a reference to an arbitrary lock, and `RawRwLock` has an unsafe method `bump_shared_slow`, but a safe method `bump_exclusive_slow`. What is the convention here, if any?

### DIFF
--- a/lock_api/src/mutex.rs
+++ b/lock_api/src/mutex.rs
@@ -273,10 +273,13 @@ impl<R: RawMutex, T: ?Sized> Mutex<R, T> {
     ///
     /// # Safety
     ///
-    /// This method is unsafe because it allows unlocking a mutex while
-    /// still holding a reference to a `MutexGuard`.
+    /// If the returned mutex is manually unlocked via [`RawMutex::unlock`], it
+    /// is your responsibility to avoid double unlocks and data races with alive
+    /// `MutexGuard`s. For example, it is sufficient to ensure that the current
+    /// thread logically owns a `MutexGuard`, but that guard has been discarded
+    /// using `mem::forget`.
     #[inline]
-    pub unsafe fn raw(&self) -> &R {
+    pub fn raw(&self) -> &R {
         &self.raw
     }
 
--- a/lock_api/src/remutex.rs
+++ b/lock_api/src/remutex.rs
@@ -378,10 +378,13 @@ impl<R: RawMutex, G: GetThreadId, T: ?Sized> ReentrantMutex<R, G, T> {
     ///
     /// # Safety
     ///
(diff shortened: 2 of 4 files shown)


Amanieu:
There is some logic to the safety of the various methods:
- The `unlock` method on the trait is unsafe because the caller needs to guarantee that the lock is currently held. This is to allow a `RawMutex` implementation to soundly access internal data in an `UnsafeCell` because it knows it has exclusive access to it.
- The implementation in `parking_lot` doesn't have any such internal data, and therefore doesn't need to use `unsafe`. In other words, if `unlock_slow` is called while the mutex isn't locked, it doesn't make our own code unsound (but it would still result in incorrect behavior, most likely leading to the caller's code being unsound).

`bump_shared_slow` being unsafe is most likely a mistake, it should be safe according to this convention.

purplesyringa:
I won't push on this anymore if you don't agree, but I want to explain my reasoning.

My intuition was that the safety requirements of the combination of `raw`/`unlock` is roughly analogous to:

```rust
let boxed = Box::new(String::new("abc"));
let x: *mut String = boxed.as_mut_ptr(); // roughly `raw`
let s = unsafe { x.read() }; // roughly `unlock`
// cannot use `boxed` safely here
unsafe { x.write(s); } // roughly re`lock` again
// can now correctly access `boxed` in safe code
```

So if `std` says `as_mut_ptr` is safe despite it enabling some unsafe operations temporarily invalidating the safe abstraction, without the documentation of `<*mut T>::write` mentioning those safe abstractions, it would be consistent for `raw` to be safe too.

In this sense, `Mutex::raw` is similar to `Mutex::data_ptr` in spirit: both ask the caller to ensure that the unsafe operations performed on the returned value do not cause data races; yet `data_ptr` is safe. Focusing on `raw` specifically just seems odd to me due to the inconsistency with both other `parking_lot` APIs and the ecosystem.

It might also be important to mention that `raw` is a method on `Mutex`, not `MutexGuard`. It doesn't seem meaningful to me to discuss whether a `MutexGuard` is accessed while the mutex is unlocked if you start with `&Mutex`. This requirement doesn't seem actionable. I'm sympathetic to enforcing safety, but `raw` just seems like a categorically wrong place to do that.

Issue #443, opened 2024-05-25
Add support for async lock functions in lock_api


Lochlanna:
With the advent of async traits in Rust, I believe it would be highly beneficial for lock_api to provide a trait and implementation that exposes asynchronous lock functions for Mutex and RwLock. Implementing versions with timeouts may be unnecessary since the futures can be cancelled by dropping them, inherently handling timeouts.

To integrate this, the new functionality would need to be conditionally compiled based on compiler support for async traits. Here's a rough outline of what I'm proposing:

```rust
#[cfg(has_async_fn_in_trait)]
pub unsafe trait RawMutexAsync: RawMutex {
    /// Locks this mutex asynchronously.
    fn lock_async(&self) -> impl Future<Output = ()> + Send;
}

#[cfg(has_async_fn_in_trait)]
impl<R: RawMutex + RawMutexAsync, T: ?Sized> Mutex<R, T> {
    pub async fn lock_async(&self) -> MutexGuard<'_, R, T> {
        RawMutexAsync::lock_async(&self.raw).await;
        // SAFETY: The lock is held, as required.
        unsafe { self.make_guard_unchecked() }
    }

    #[cfg(feature = "arc_lock")]
    pub async fn lock_async_arc(self: &Arc<Self>) -> ArcMutexGuard<R, T> {
        RawMutexAsync::lock_async(&self.raw).await;
        // SAFETY: The lock is held, as required.
        unsafe { self.make_arc_guard_unchecked() }
    }
}
```

A potential issue is the reliance on `RawMutex` being implemented as well. This approach wouldn't work for purely async locks. One possible workaround could be to split the `RawMutex` trait into two traits: one containing the shared items and another without them. However, this would be a significant change.

I'd love to hear your thoughts on this.

Amanieu:
My position has always been that async locks need a completely different interface and implementation than normal lock, and therefore should be in a separate crate from parking_lot and lock_api.

Lochlanna:
Makes sense. I'm trying to figure out the best way to setup my lock which is a hybrid that supports both sync and async at the same time. I'll probably just roll a custom API for it.

Issue #86, opened 2018-08-16
Async handling


Ekleog:
This is a follow-up from https://github.com/Amanieu/parking_lot/issues/24, hopefully more explicit :)

So first, a disclaimer: I know literally nothing about the internals of `parking_lot`, and very little about mutexes.

What I'd be looking for is a way to have `parking_lot` mutexes play nicely with futures. What I mean by “play nicely” is:
 * It should never block the whole thread, only a task on the thread pool
 * It should return eg. a `Future<Item = MutexGuard<'a, T>>`

The second point can easily be implemented by wrapping `try_lock`, so that leaves the first point to handle.

My reflection on `parking_lot` internals (please tell me if it's wrong) is:
 * There is a “low-level” layer, which puts a thread to sleep, wake up a thread, etc.
 * Upon these building blocks, mutexes are built
 * Upon mutexes, guards are built

Most of the complexity would stand in the second layer above.

What I'm thinking is that, if the first layer could be swapped with an async-friendly layer (that would put a task to sleep instead of the whole thread, wake up a task, etc.), then I think it would become possible to reuse the biggest part of the code, that is in how mutexes are built on top of these primitives.

A way to do that may be to parameterize all `parking_lot` types on a helper trait that provides said functions, and static-dispatch on it, with type aliases for more convenience.

Now, as stated above, I know nothing about the internals of `parking_lot`, so I may completely be wrong in my analysis. Feel free to tell it to me if so :)

So, what do you think about this idea of splitting away the low-level task handling from the mutex building?

Amanieu:
I don't think that simply parameterizing the lock types is sufficient: what you are suggesting (changing the return type from `MutexGuard` to `Future<Item = MutexGuard>`)  is actually a pretty significant API change.

The `lock_api` crate is designed to provide a high-level API on top of a basic raw mutex type (the top-most level from your description). While it might be possible to add support for futures to this, it would introduce a lot of complexity to what is already a very complex crate.

A bigger issue is that the mechanism used by `Future` to wake up a thread is very simple: you just call `wake()` on a `Waker` object that you obtain when the `Future` is polled. The `parking_lot_core` API is  much more advanced and allows for things like fairness guarantees, passing in tokens when parking/unparking, and callback which allow fine-grained control of how threads are woken up. The mutex algorithm in `parking_lot` (the middle level from your description) heavily relies on these features and most likely won't work without them.

[...]

PR #517, opened 2026-02-08
Support for asynchronous parking

Wulf0x67E7:
# Summary
Added support for asynchronous parking to parking_lot_core behind a feature flag, and in the process also made all the (un)park-methods reuse their queue manipulation logic, as well as added a future extension trait and methods to util.
# Motivation
I was recently playing around with synchronization primitives and sync/async interop. While trying to implement [something](https://github.com/Wulf0x67E7/conutex) using parking_lot_core, I wanted to add async support, but realized that parking lot didn't support what I needed for that. I then thought that extending it myself would be a fun and doable project.
# Changes
 - Added feature "async"

 - Added optional dependency "futures" (couldn't work out how to make it a feature dependent dev-dependency)
 - Added function `parking_lot_core::park_task`
 - Added `ImmediateFuture` utility trait, struct, and functions
 - Refactored `park` and `park_task` to reuse shared queue manipulation logic
 - Refactored the various `unpark` methods to also reuse the shared queue manipulation logic
 - Extended the testing tools to also work with async and mixed sync/async operations and made it pass both normally and under miri
 - Added documentation and comments for the new/refactored code
 - [025af7e] Added dependency `scopeguard = { version = "1.1.0", default-features = false }`
# Issues
 - Full optional "futures" dependency instead of just dev-dependency
 - [025af7e] ~`park_task` has even stricter safety requirements then `park`, namely that the resulting future can not be dropped or forgotten before being unparked or timed out.~
 - [025af7e] ~Because of that, when using `park_task` to implement e.g. an asynchronous-aware Mutex its unsafety cannot be fully mitigated unless you yourself are the one synchronously polling it to completion, which kinda defeats the point.~
# Notes
 - [025af7e] Leaking the future from `park_task`~, unlike dropping or forgetting,~ is actually fine and won't even block other things from making progress because `ImmediateFuture` ensures it cant yield while still holding the queue lock.
 - [025af7e]~Thinking about it now, leaking `park_task`s `ParkData` at the beginning and then only "un-leaking"  and dropping it at the end could fix the extra unsafety, but would require allocation and, obviously, risk leaking memory.~
 - The aproach used in [025af7e]  to make `park_task` panic safe could also be applied to `park`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 target
 Cargo.lock
+.vscode
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -29,5 +29,7 @@ redox_syscall = "0.5"
 windows-link = "0.2.0"
 
 [features]
+default = ["async"]
 nightly = []
+async = []
 deadlock_detection = ["petgraph", "backtrace"]
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -18,6 +18,7 @@ cfg-if = "1.0.0"
 smallvec = "1.6.1"
 petgraph = { version = "0.6.0", optional = true }
 backtrace = { version = "0.3.60", optional = true }
+futures = { version = "0.3.31", optional = true }
 
 [target.'cfg(unix)'.dependencies]
 libc = "0.2.95"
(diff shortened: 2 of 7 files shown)


faern:
Highly related issues: #443, #86. Async support in parking_lot has been discussed before and closed as not in scope at least those times.

Issue #440, opened 2024-04-26
[Feature Request] Create `OnceLock` and `OnceCell` using this library's Once


BLucky-gh:
Since unlike `Mutex`, you can't specify which underlying `Once` to use with `std::sync::OnceLock` and `std::cell::OnceCell`, I would either have to roll my own, which seems wasteful and will not update when the std versions get changed, or manage the value and the `Once` separately like in the `Once` example, which is error-prone and puts the burden of upholding invariants on the developer

Amanieu:
I would be happy to accept a PR for it, but it is not something I have time to work on myself.

BLucky-gh:
I could try making a PR for it, but it will probably be almost entirely just a copy paste of the `std` implementation and just changing the import from `std::sync::Once` to `parking_lot::Once`, would that be any good?

Amanieu:
That's fine.

PR #519, opened 2026-04-11
feat: implement `OnceLock`

chise0713:
This PR is based on the `OnceLock` implementation from the standard library, with modifications to adapt it to our `Once` implementation.

Related: #440

Tests may benefit from further validation and edge case coverage.

Any feedback on the design and implementation would be appreciated.


src/once.rs
```diff
@@ -196,7 +226,12 @@ impl Once {
     // currently no way to take an `FnOnce` and call it via virtual dispatch
     // without some allocation overhead.
     #[cold]
-    fn call_once_slow(&self, ignore_poison: bool, f: &mut dyn FnMut(OnceState)) {
+    fn call_once_slow(
+        &self,
+        ignore_poison: bool,
+        from_once_lock: bool, // work around for `OnceLock::get_or_try_init`
```
chise0713: To fix incorrect `Err` handling in `OnceLock::get_or_try_init`
and align with std semantics, I extended `Once`'s internal API
to allow manual control over its state.

This keeps the public API unchanged, but adds a `OnceLock`-specific
code path internally.

I'm not entirely confident this is the best approach, so I'd really
appreciate any feedback!

src/once_lock.rs
```diff
@@ -0,0 +1,858 @@
...
+///         42
+///     })
+/// }
+/// ```
+pub struct OnceLock<T> {
+    once: Once,
+    value: UnsafeCell<MaybeUninit<T>>,
+}
+
+impl<T> OnceLock<T> {
```
chise0713, line 43: The documentation is mostly adapted from the standard library.

PR #522, opened 2026-07-28
Remove redox implementation

willnode:
Redox OS has been `cfg(unix)` [for years](https://github.com/rust-lang/rust/issues/60139) so the specialized code here can be considered a dead code.

Checked to compile on Redox target but the tests is hanging, I suppose it's because of relibc implementation not related with this PR.


Amanieu:
Should we instead be keeping this code and just preferring it over the generic unix version on redox? I would like to avoid the pthreads API whenever possible.

willnode:
Redox (relibc) pthread is a wrapper to kernel futex, so it won't go away. I think you have a good point, maybe doing that could make tests passing in Redox platform. 

The downside to that approach is that `redox_syscall` crate is "unstable", so the crate version kept bumping up, right now it's at `"0.9.1"` and maybe it will get bumped again. It's a minor problem though, because futex is fairly stable even today. But i'll ask redox dev maybe we can get some better feedback.

Amanieu:
If you compare the unix and linux thread parkers then you will see that the unix one is much more complicated because of the pthreads API. In particular it requires run-time initialized and a drop, which none of the other thread parkers need.

But in the end, if redox doesn't guarantee a stable syscall API then we have no choice but to use the stable pthreads interface.

willnode:
Redox does have a stable API on [libredox crate](https://crates.io/crates/libredox/0.1.19), but there seems no API related to futex or any other thread parking interface. Adding futex API there should be fairly easy, but Redox is focusing on getting new release version right now so it will take time to make it appear.

In the mean time, keep using `cfg(unix)` is best option for now, so maybe we can keep the redox code intact until `libredox` have thread parking API?

Amanieu:
Yes, let's keep the code around for now.

willnode:
Closed, i'll get back with libredox in future, thanks!

PR #523, opened 2026-07-29
Move `Condvar` into `lock_api`

bushrat011899:
# Background

While working on [gfx-rs/wgpu#9780](https://github.com/gfx-rs/wgpu/pull/9780) I was surprised to see that `Condvar` is only available through `parking_lot`, unlike `Mutex` which is available in `lock_api`. This was unfortunate since that PR's objective was to use `lock_api` to abstract over `no_std` vs `std` enabled locking primitives.

## Objective

* Move `parking_lot::Condvar` to `lock_api` with a type parameter to allow for an arbitrary `RawMutex` to be provided.

### `lock_api`

* Moved `WaitTimeoutResult` unmodified from `parking_lot` to `lock_api`.
* Added a new unsafe trait `RawCondvar` which provides the implementation specific functionality of a condvar. An associated type parameter is used to prescribe the `RawMutex` type this condvar is compatible with.
* Added another unsafe trait `RawCondvarTimed` which adds timeout support to `RawCondvar`.
* Added a struct `Condvar<C>` which provides an identical API to `parking_lot::Condvar` for any `C: RawCondvarTimed`, and a subset for `C: RawCondvar`.
* Verified existing MSRV of 1.71.0 is sufficient.

### `parking_lot`

* Changed `Condvar` to an implementation of `RawCondvarTimed` and renamed to `RawCondvar`.
* Reexported `lock_api::WaitTimeoutResult` and `lock_api::Condvar` with `C = parking_lot::RawCondvar`.
* Adjusted tests for slightly altered (internal) API.

---

## Notes

* No AI tooling of any kind was used during the creation of this PR.
* Happy to open an issue if requested, as I noticed this doesn't appear to have been requested before.


lock_api/src/condvar.rs
```diff
@@ -0,0 +1,410 @@
...
+    /// If there is a blocked thread on this condition variable, then it will
+    /// be woken up from its call to `wait` or `wait_timeout`. Calls to
+    /// `notify_one` are not buffered in any way.
+    ///
+    /// To wake up all threads, see `notify_all()`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use parking_lot::Condvar;
```
bushrat011899: Forgot to update the doc examples after moving.

bushrat011899: Removed examples (only a couple) and updated documentation to remove implementation details that are specific to `parking_lot`. Not sure if those details should be moved onto the `RawCondvar` implementation itself, so noting here for visibility.

Amanieu:
This mostly looks good! I only have one concern: `RawCondvar::wait` should be clearer in its documentation about what should happen if you use different `RawMutex` simultaneously with a a single condvar. There is a wide spectrum of how different implementations support this:
- In parking_lot this panics, but only if there are 2 concurrent waits with different mutexes.
- In std's futex-based implementation this just works.
- In pthreads this is UB.
- In std's pthreads-based implementation this panics if you ever use 2 different mutexes with the same condvar.

I would say that this should be documented in the trait as:
- May panic if used with 2 different mutexes.
- Must not cause UB if used with 2 different mutexes.

bushrat011899:
Excellent to hear! I've added safety requirements to both new unsafe traits, and documented on user facing methods that the underlying implementation is permitted to panic.

lock_api/src/condvar.rs
```diff
@@ -0,0 +1,386 @@
...
+    /// The type of [`RawMutex`] this condvar can work with.
+    type RawMutex: RawMutex;
+
+    /// Wait until the provided [`RawMutex`] is available.
+    ///
+    /// # Panics
+    ///
+    /// Implementations are permitted to panic if requested to wait on two distinct
+    /// [`RawMutex`]s simultaneously.
+    fn wait(&self, mutex: &Self::RawMutex);
```
Amanieu: This and all the other `wait` methods should be unsafe for the same reason that `RawMutex::unlock` is unsafe: there is a safety invariant that the caller has locked the mutex and owns the lock for the duration of the call.

lock_api/src/condvar.rs
```diff
@@ -0,0 +1,386 @@
...
+        F: FnMut(&mut T) -> bool,
+    {
+        let mut result = WaitTimeoutResult(false);
+
+        while !result.timed_out() && condition(mutex_guard.deref_mut()) {
+            result = WaitTimeoutResult(
+                self.inner
+                    .wait_for(unsafe { MutexGuard::mutex(mutex_guard).raw() }, &timeout),
+            );
+        }
```
Amanieu, line 408: This doesn't match the old behavior: we want to convert the duration to a deadline first, otherwise each inner wait in the loop gets a full duration timeout every time the inner `wait_for` is called.

The best way to handle this is to add a `duration_to_deadline` associated function to `RawCondvarTimed` to convert the timeout to a deadline at the start of the loop and then using wait_until internally.

I would keep `RawCondvarTimed::wait_for` on the trait because some implementations may only have native support for timeouts, not deadlines. But it should have a default body that calls `duration_to_deadline` and then calls `wait_until`.

Amanieu (commented):
Some additional comments came up from a second round of review.

lock_api/src/condvar.rs
bushrat011899: Oversight on my part, good catch!

lock_api/src/condvar.rs
bushrat011899: I've added a `checked_duration_to_instant` since the current implementation uses `Instant::checked_add`, and I want to preserve that. From what I can tell, when that checked conversion fails, the method effectively degrades from a `wait_for_until` into a `wait_until` (and likewise for `wait_for`/`wait`).

bushrat011899:
I believe I've addressed both concerns. I've also squashed the commits up until this point since this latest round was mostly just changing the first commit, and I like to keep commits as atomic as possible for `git bisect`.

PR #524, opened 2026-07-29
Add `lock_api::{Lock}::into_inner_with_raw`

### LINKED ISSUE #521: lock_api::Mutex: Get original raw mutex when destructuring
Hello,

I'm using lock_api in embedded development.

First I was hitting the same problem as in issue #400, which is now solved by `Mutex::from_raw`.

Now new problem arose - when using Mutex in typestate-pattern structs, I would often need to destructure the Mutex into original raw mutex, and the data.

Existing method `into_inner` only destructures into data.

So I suggest adding method `fn Lock::into_raw_inner(self) -> (R, T)`.

I can make a PR if we agree on the API.

bushrat011899:
# Objective

- Fix #521 with the [agreed upon](https://github.com/Amanieu/parking_lot/issues/521#issuecomment-4654193357) solution.

### `lock_api`

- Added `{Mutex, ReentrantMutex, RwLock}::into_inner_with_raw`, which returns the values required to call the relevant `from_raw` function.

---

## Notes

- No AI tooling of any kind was used during the creation of this PR.

### DIFF
--- a/lock_api/src/remutex.rs
+++ b/lock_api/src/remutex.rs
@@ -276,6 +276,17 @@ impl<R, G, T> ReentrantMutex<R, G, T> {
     pub const fn const_new(raw_mutex: R, get_thread_id: G, val: T) -> ReentrantMutex<R, G, T> {
         Self::from_raw(raw_mutex, get_thread_id, val)
     }
+
+    /// Consumes this mutex, returning the underlying data, raw mutex and
+    /// thread ID helper.
+    #[inline]
+    pub fn into_inner_with_raw(self) -> (R, G, T) {
+        (
+            self.raw.mutex,
+            self.raw.get_thread_id,
+            self.data.into_inner(),
+        )
+    }
 }
 
 impl<R: RawMutex, G: GetThreadId, T: ?Sized> ReentrantMutex<R, G, T> {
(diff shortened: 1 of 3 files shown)


bushrat011899:
I believe the MacOS CI failure is spurious, since it failed during setup.

lock_api/src/remutex.rs
```diff
@@ -276,6 +276,17 @@ impl<R, G, T> ReentrantMutex<R, G, T> {
     pub const fn const_new(raw_mutex: R, get_thread_id: G, val: T) -> ReentrantMutex<R, G, T> {
         Self::from_raw(raw_mutex, get_thread_id, val)
     }
+
+    /// Consumes this mutex, returning the underlying data and raw mutex and
+    /// thread ID helper.
```
Amanieu:
```suggestion
    /// Consumes this mutex, returning the underlying data, raw mutex and
    /// thread ID helper.
```

Amanieu:
LGTM apart from a small grammar nit.

lock_api/src/remutex.rs
bushrat011899: Amended the existing commit, good catch!

PR #525, opened 2026-07-29
Split `RawMutex` and `RawRwLock` into core and init traits

### LINKED ISSUE #117: lock_api and abstracting over Mutexes which require initialization
I'm currently working on a library which provides some higher level synchronized data-structures, and aims to work on `std` environments as well as on bare-metal and other environments.
The main thing that is not in common between those environments are Mutexes for my case, so I would like to be able to abstract over those.

I tried building a custom trait first, but without GATs it doesn't seem to work well with the associated data and lockguards with lifetimes. Now I discovered `lock_api`, which seems to roughly do what I want - making it possible to define predefined data-types which make use of parking_lot mutexes, and allow the users to override e.g. to a `SpinLock` using a template parameter.

However I'm wondering about the support for some platforms, where the following applies:
Some `RawMutex` types require an explicit call for initialization, e.g. native system mutexes on lots of platforms from Windows to FreeRTOS. They require to call some kind of `CreateMutex()` call, which may fail. This seems to not be covered by lock_api, since the `RawMutex` trait accepts only a const init value instead of a `new()` function. I guess this is intentional, and not subject for change?

How would we support using `lock_api` based types on top of a platform like FreeRTOS? Would we port parking_lot to it? Or maybe set `INIT` to `0` and try to initialize the mutex on the first call to `lock()` within a global critical section?
Those init functions can obviously also fail, so `new()` could be allowed to return an error, which however would then also require `Mutex::new()` to return an error. It probably might not make sense to support this, since it would clutter up the APIs for all other platforms, and the best practice should be not to allocate those kinds of resources after startup. So `panic!`ing might be ok.

And an unrelated question: How stable is the `lock_api` API? It seems like all the basics should be covered. Yet it still sits only at a very low version number of 0.1.5?

### LINKED ISSUE #400: INIT requirement constrains which types can be used to implement RawMutex
I think the lock-api crate has a lot of potential in embedded, but right now it has a limitation that implementations of `RawMutex` must provide an `INIT`  constant value.

Some MCUs have their own hardware synchronization primitives, but these are unique resources / peripherals that must be owned in order to use them, and ownership is generally handed to the application through a PAC, or "peripheral access crate". They naturally cannot be obtained in a const context, and they also cannot be cloned since they rely on ownership being unique.

Peripherals like [`rp2040_hal::sio::Spinlock`](https://docs.rs/rp2040-hal/0.9.0/rp2040_hal/sio/struct.Spinlock.html) would be perfect for implementing `RawMutex`, but it's impossible to create a value for `INIT` without some kind of wrapper type that provides an illegal state.

`INIT` only seems necessary for `Mutex::new` and similar constructors that do not explicitly accept a `RawMutex` parameter, so I think it would make sense to move it into a separate trait. The raw mutex can still be used with the `const_new` constructors; it would operate similar to constructors that accept an allocator argument (`Box::new_in` etc.).

bushrat011899:
# Objective

- Fix #117
- Fix #400

## Solution

- Solution based on previously approved suggestion [here](https://github.com/Amanieu/parking_lot/issues/400#issuecomment-1707454078).
- Created a new pair of traits which supersede `RawMutex`: `RawMutexCore` and `RawMutexInit`. `RawMutex` is left unchanged, and the new traits are blanket implemented for any type implementing this original trait. This allows the split to be a non-breaking change.
- Created a new pair of traits which supersede `RawRwLock`: `RawRwLockCore` and `RawRwLockInit`. The same blanket implementation as used for `RawMutex` is also used for `RawRwLock`.
- Replaced trait bounds across `parking_lot` to refer to core and init traits as required.
- Ensured `parking_lot` and other downstream crates observe no breakage from this change.

---

## Notes

- No AI tooling of any kind was used during the creation of this PR.
- `RawRwLock` and `RawMutex` could be deprecated, or removed entirely (justifying renaming `RawMutexCore` to `RawMutex`). I have not done this, but I'd be happy to amend the PR to do so.

### DIFF
--- a/lock_api/src/mutex.rs
+++ b/lock_api/src/mutex.rs
@@ -33,6 +33,89 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
 ///
 /// Implementations of this trait must ensure that the mutex is actually
 /// exclusive: a lock can't be acquired while the mutex is already locked.
+pub unsafe trait RawMutexCore {
+    /// Marker type which determines whether a lock guard should be `Send`. Use
+    /// one of the `GuardSend` or `GuardNoSend` helper types here.
+    type GuardMarker;
+
+    /// Acquires this mutex, blocking the current thread until it is able to do so.
+    fn lock(&self);
+
+    /// Attempts to acquire this mutex without blocking. Returns `true`
+    /// if the lock was successfully acquired and `false` otherwise.
+    fn try_lock(&self) -> bool;
+
+    /// Unlocks this mutex.
+    ///
+    /// # Safety
+    ///
+    /// This method may only be called if the mutex is held in the current context, i.e. it must
+    /// be paired with a successful call to [`lock`], [`try_lock`], [`try_lock_for`] or [`try_lock_until`].
+    ///
+    /// [`lock`]: RawMutexCore::lock
+    /// [`try_lock`]: RawMutexCore::try_lock
+    /// [`try_lock_for`]: RawMutexTimed::try_lock_for
+    /// [`try_lock_until`]: RawMutexTimed::try_lock_until
+    unsafe fn unlock(&self);
+
+    /// Checks whether the mutex is currently locked.
+    #[inline]
+    fn is_locked(&self) -> bool {
+        let acquired_lock = self.try_lock();
+        if acquired_lock {
+            // Safety: The lock has been successfully acquired above.
+            unsafe {
+                self.unlock();
+            }
+        }
+        !acquired_lock
+    }
+}
+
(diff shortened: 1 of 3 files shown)


Amanieu:
I'm not a huge fan of the amount of churn here, especially considering the fact that in the vast majority of cases the const initializer can just be used. In the long term I want to wait until we have const traits and then just require `~const Default` in `Mutex::new` which can be const if the raw mutex type has a const `Default` impl.

For now I prefer just requiring users who need non-const-constructible mutexes to just use `INIT = unimplemented!()` and `from_raw`.

bushrat011899:
Understandable! I'll close this PR, thanks for the quick feedback!

PR #529, opened 2026-08-12
Add a stable thread parker for threaded wasi targets

michaelkremenetsky:
## Problem

On `wasm32-wasip1-threads`, parking_lot only has a working thread parker if you build with the `nightly` feature: the `wasm_atomic` parker needs the `memory.atomic.wait32`/`notify` intrinsics, which are still unstable (rust-lang/rust#77839, and stabilization looks stuck on bigger design questions about the `atomics` target feature). On stable, the cfg chain falls through to `wasm.rs`, which panics with `"Parking not supported on this platform"` the first time anything actually parks.

This bites in practice: napi-rs wasi npm packages (rolldown, `@tailwindcss/oxide`, and friends) are built on stable and shipped prebuilt, so the panicking stub is baked into binaries on npm today. They work until a lock is contended or a condvar waits, then abort. The current workarounds are keeping everything single-threaded or requiring a nightly toolchain for every binding author, neither of which is great.

## What this does

Adds `wasi_threads.rs`, a parker built on `std::sync::Mutex` + `Condvar`, which are futex-backed on threaded wasi targets, and selects it for `all(target_family = "wasm", target_os = "wasi", target_feature = "atomics")` when the `nightly` parker isn't enabled.

The rest of the selection order is unchanged. `nightly` + atomics still gets `wasm_atomic`, which stays the preferred parker (this one goes through std's Mutex/Condvar, so it's slower than a raw futex wait). Non-threaded wasm keeps the stub, and `wasm32-unknown-unknown` is untouched since the new cfg requires `target_os = "wasi"`.

Happy to rename the file or adjust the cfg (e.g. if you'd rather scope it to `wasip1-threads` specifically).


Amanieu:
Unfortunately this won't work as you expect: the `atomics` feature is still unstable, which means that it only appears as available when using a nightly compiler. On a stable compiler the `atomics` feature will be unavailable and it will fall back to the default wasm parker.

I would encourage you to instead push towards the stabilization of WASM atomics in Rust by commenting on https://github.com/rust-lang/rust/issues/77839.

PR #530, opened 2026-08-14
Tighten Mapped guard Send/Sync bounds to require R: Sync

### LINKED ISSUE #527: lock_api: the mapped guards' `Send`/`Sync` impls omit `R: Sync`
## What this is and is not

This is **a residual gap in the RUSTSEC-2020-0070 fix, not a new class of bug**, and it is worth
stating that plainly up front rather than at the bottom. The 2020 fix corrected the mapped guards'
marker impls on the **data** type parameter. The same impls were left unbounded on the **raw lock**
type parameter, and that is still the case in 0.4.14.

It is **latent**. It requires a downstream crate to implement `RawMutex` or `RawRwLock` for a type
that is `Send` but not `Sync`. Every raw lock in wide use, `parking_lot`'s included, is `Sync`, so
there is no known affected shipping code and no reason to think anyone has hit this. I am not
requesting an advisory and not claiming a severity; the appropriate fix is a bound tightening.

It is **not a 0.4.14 regression**. The complete set of `unsafe impl` lines is byte-for-byte
identical between 0.4.13 and 0.4.14, so the shape predates both and most likely dates to the 2020
fix itself.

## The defect

Each mapped guard stores a borrow of the raw lock:

```rust
pub struct MappedMutexGuard<'a, R: RawMutex, T: ?Sized> {
    raw: &'a R,                       // mutex.rs:878
    data: *mut T,
    marker: PhantomData<&'a mut T>,
}
```

and its `Drop` calls an `&self` method on it:

```rust
impl<'a, R: RawMutex + 'a, T: ?Sized + 'a> Drop for MappedMutexGuard<'a, R, T> {
    fn drop(&mut self) { unsafe { self.raw.unlock(); } }        // mutex.rs:1019
}
```

Moving or sharing a value that holds `&'a R` across a thread boundary requires `R: Sync` — that is
the definition of `Sync`. Five hand-written impls do not require it:

| file:line | impl |
|---|---|
| `mutex.rs:887` | `unsafe impl<'a, R: RawMutex + 'a, T: ?Sized + Send + 'a> Send for MappedMutexGuard<'a, R, T> where R::GuardMarker: Send` |
| `rwlock.rs:2801` | `unsafe impl<'a, R: RawRwLock + 'a, T: ?Sized + Sync + 'a> Sync for MappedRwLockReadGuard<'a, R, T> {}` |
| `rwlock.rs:2802` | `unsafe impl<'a, R: RawRwLock + 'a, T: ?Sized + Sync + 'a> Send for MappedRwLockReadGuard<'a, R, T> where R::GuardMarker: Send` |
| `rwlock.rs:2969` | `unsafe impl<'a, R: RawRwLock + 'a, T: ?Sized + Sync + 'a> Sync for MappedRwLockWriteGuard<'a, R, T>` |
| `rwlock.rs:2973` | `unsafe impl<'a, R: RawRwLock + 'a, T: ?Sized + Send + 'a> Send for MappedRwLockWriteGuard<'a, R, T> where R::GuardMarker: Send` |

The corresponding `Drop` impls are `mutex.rs:1019` (`raw.unlock()`), `rwlock.rs:2927`
(`raw.unlock_shared()`) and `rwlock.rs:3105` (`raw.unlock_exclusive()`).

For the three `Send` impls the omission is the whole story, because each mapped guard carries a
`*mut T` or `*const T` data field. That already makes the auto impl negative, so the `unsafe impl`
is not *relaxing* a bound — it is what **grants** `Send` in the first place, and its bound list is
the only thing between a caller and a cross-thread `&R`.

## Why this reads as an oversight rather than a decision

The inconsistency is internal to the crate, and in one case internal to a single type:

- **`MappedMutexGuard`'s own `Sync` impl carries the bound**, four lines above the `Send` impl that
  does not:
  ```rust
  unsafe impl<'a, R: RawMutex + Sync + 'a, T: ?Sized + Sync + 'a> Sync for MappedMutexGuard<'a, R, T> {}   // mutex.rs:883
  unsafe impl<'a, R: RawMutex + 'a,        T: ?Sized + Send + 'a> Send for MappedMutexGuard<'a, R, T>      // mutex.rs:887
      where R::GuardMarker: Send {}
  ```
- **`ArcMutexGuard`'s `Send` spells it out**: `R: RawMutex + Send + Sync` (`mutex.rs:739`).
- **`MutexGuard`, `RwLockReadGuard` and `RwLockWriteGuard` hand-write no `Send` at all**, so they
  acquire the requirement transitively and correctly: `&'a Mutex<R, T>: Send` needs
  `Mutex<R, T>: Sync`, which `mutex.rs:144` defines as `R: RawMutex + Sync, T: Send`. Same shape at
  `rwlock.rs:365` for `RwLock`.
- The crate writes `R: Sync` explicitly in **twelve** places — `mutex.rs:144, 509, 739, 744, 883`;
  `remutex.rs:72, 225, 623, 978`; `rwlock.rs:365, 1258, 1602`.

That last point disposes of the obvious objection, which is worth answering directly since it is
the strongest one available: *"`RawMutex` is a synchronization primitive, so an implementor is
implicitly required to be `Sync`, and a `!Sync` implementor is the caller's fault."* If that were
the crate's model, `unsafe impl<R: RawMutex + Sync, ...> Sync for Mutex<R, T>` would be redundant,
and so would the other eleven. The documented safety contract requires only exclusivity:

> Implementations of this trait must ensure that the mutex is actually exclusive: a lock can't be
> acquired while the mutex is already locked. — `mutex.rs:32-35`

It says nothing about `Sync` or thread-safety. The crate therefore treats a `Send`-but-`!Sync` raw
lock as a supported configuration confined to one thread, and the mapped guards are what let it
escape that thread.

## Witness

The obvious witness is unsound in itself and should not be used: a `RawMutex` whose *exclusion
mechanism* is a `Cell<bool>` cannot deliver mutual exclusion across threads, so it breaks the
`mutex.rs:32-35` contract and the resulting UB is fairly charged to the implementor. The witness
below keeps exclusion **entirely atomic**, so the contract is fully upheld, and makes the
`!Sync`-ness orthogonal, non-load-bearing bookkeeping:

```rust
use core::cell::Cell;
use core::sync::atomic::{AtomicBool, Ordering};
use lock_api::{GuardSend, Mutex, MutexGuard, RawMutex};

pub struct CountingRaw {
    locked: AtomicBool,
    // Orthogonal bookkeeping. Makes the type !Sync without weakening exclusivity.
    events: Cell<usize>,
}

unsafe impl RawMutex for CountingRaw {
    const INIT: CountingRaw = CountingRaw {
        locked: AtomicBool::new(false),
        events: Cell::new(0),
    };
    type GuardMarker = GuardSend;

    fn lock(&self) { while !self.try_lock() {} }

    fn try_lock(&self) -> bool {
        let got = self
            .locked
            .compare_exchange(false, true, Ordering::Acquire, Ordering::Relaxed)
            .is_ok();
        self.events.set(self.events.get() + 1);   // non-atomic, and NOT under the lock
        got
    }

    unsafe fn unlock(&self) {
        self.events.set(self.events.get() + 1);   // non-atomic
        self.locked.store(false, Ordering::Release);
    }
}

fn main() {
    // 'static so plain thread::spawn applies; nothing here is unsafe.
    let m: &'static Mutex<CountingRaw, i32> = Box::leak(Box::new(Mutex::new(0)));

    let guard = m.lock();
    let mapped = MutexGuard::map(guard, |v| v);   // MappedMutexGuard<'static, CountingRaw, i32>

    std::thread::spawn(move || {
        drop(mapped);            // Drop -> raw.unlock() -> writes `events` on this thread
    });

    loop {
        let _ = m.is_locked();   // -> try_lock() -> writes `events` on the main thread
    }
}
```

`CountingRaw` is `Send` (an `AtomicBool` and a `Cell<usize>` are both `Send`) and `!Sync` (the
`Cell` is not). The `spawn` is accepted purely by `mutex.rs:887`: it needs `R: RawMutex`,
`T: Send`, `R::GuardMarker: Send`, and asks nothing about `R: Sync`. Two threads then
read-modify-write the same non-atomic `Cell<usize>` with no synchronization between them.

The contrast is the clearest statement of the problem: **`MutexGuard` itself cannot be sent this
way.** Its `Send` is the auto impl, which needs `&'static Mutex<CountingRaw, i32>: Send`, hence
`Mutex<CountingRaw, i32>: Sync`, hence `CountingRaw: Sync` — which does not hold, so the compiler
rejects it. `MutexGuard::map` turns a guard that cannot cross a thread boundary into one that can,
and the bound that was doing the work is not carried across.

## Relationship to RUSTSEC-2020-0070

RUSTSEC-2020-0070 (`CVE-2020-35910` .. `-35914`, parking_lot PR #262, `informational = "unsound"`,
patched `>= 0.4.2`) lists exactly `lock_api::MappedMutexGuard`,
`lock_api::MappedRwLockReadGuard`, `lock_api::MappedRwLockWriteGuard`,
`lock_api::RwLockReadGuard` and `lock_api::RwLockWriteGuard`, and describes the fix as:

> This issue was fixed by changing the trait bounds on the `Mapped` guard types and removing the
> `Sync` trait for the `RwLock` guards.

So the fix operated on the **data** parameter — which is why `rwlock.rs:2802`'s `Send` impl carries
the otherwise-surprising `T: Sync` for a read guard — and left `R` unbounded. The `Sync` impls on
the plain `RwLock` guards were later reintroduced *with* `R: Sync` (`rwlock.rs:1258`, `1602`),
which is correct and is part of why the omission on the mapped guards stands out.

Because 0.4.14 is inside the advisory's patched range, no advisory-checking tool flags this;
`cargo deny check advisories` reports `advisories ok` against a graph containing 0.4.14.

## Suggested fix

Add `+ Sync` to `R` in the five impls above. Concretely, for `mutex.rs:887`:

```rust
unsafe impl<'a, R: RawMutex + Sync + 'a, T: ?Sized + Send + 'a> Send for MappedMutexGuard<'a, R, T>
    where R::GuardMarker: Send {}
```

and the same addition at `rwlock.rs:2801`, `2802`, `2969`, `2973`. This matches what
`mutex.rs:883` and `mutex.rs:739` already do, so it makes the crate internally consistent rather
than introducing a new convention.

It is a breaking change **only** for a downstream crate that today sends a mapped guard derived
from a `!Sync` raw lock — which is precisely the unsound case.

## Adjacent, listed separately because the mechanism differs

`rwlock.rs:2052` — `Sync for RwLockUpgradableReadGuard` — also omits `R: Sync`, where the read and
write guards carry it (`rwlock.rs:1258`, `1602`):

```rust
unsafe impl<'a, R: RawRwLockUpgrade + 'a, T: ?Sized + Sync + 'a> Sync
    for RwLockUpgradableReadGuard<'a, R, T> {}
```

This guard holds `rwlock: &'a RwLock<R, T>` (`rwlock.rs:2048`) rather than `&'a R`, and
`RwLock::raw()` is `unsafe`, so the route to `R` is not a direct reference. It is instead
`RwLockUpgradableReadGuard::rwlock()` (`rwlock.rs:2059`, safe, returns `&'a RwLock<R, T>`) followed
by any safe method on `&RwLock` — `is_locked`, `try_read`, and so on — each of which calls
`self.raw.<method>()`. Reviewers of this finding split on whether to count it with the other five.
It is listed apart because the mechanism is different, not because it is weaker. Note that this
impl also omits the `T: Send` half of `RwLock`'s own `Sync` bound (`rwlock.rs:365`).

## Provenance

Found while certifying `lock_api 0.4.14` for warden's `cargo vet` store. Single reader for the
5,425-line read; the finding and its two reachability premises were then given to four independent
reviewers instructed to refute by default and to work from the crate source rather than from this
text. None refuted it. Their corrections are in this document: the site count moved from three to
five, the witness was replaced for the reason given above, and the framing changed from "novel" to
"residual of a previously-advised fix". Two incidental counts in the first write-up were wrong and
are corrected here.

## Upstream context

Checked against the default branch before filing: `lock_api/src/mutex.rs:887` still reads
`unsafe impl<'a, R: RawMutex + 'a, T: ?Sized + Send + 'a> Send for MappedMutexGuard<'a, R, T>`,
so the omission described above is current, not a stale read of 0.4.14.

Deliberately cross-referenced so this is not mistaken for a re-report of settled work:
#258 (`Mapped*Guard` objects should only be `Send` when the contained type is) and #259
(`RwLockReadGuard`/`RwLockWriteGuard` should only be `Sync` when the `RawRwLock` is) are the
issues behind RUSTSEC-2020-0070. Both are fixed. This report is about what those fixes did not
reach: they bounded the **data** parameter on the mapped guards, and #259's `R: Sync` bound was
applied to the **plain** guards (`rwlock.rs:1258`, `1602`) but not to the **mapped** ones. That
is the entire content of this issue.

MsfPablo:
Closes #527

The unsafe impls of Send/Sync for the mapped guard types only require the guarded data type to be Send/Sync, but the Drop impl of a mapped guard unlocks the underlying RawMutex / RawRwLock. A Send-but-!Sync raw mutex can carry thread-local bookkeeping; unlocking it from another thread is UB.

This commit adds R: Sync to the Sync impl and R: Send to the Send impl, matching the same tightening that landed in 2020 for the data-bound versions of the guards.

Disclosed: implementation, commit message, and PR text were prepared with AI assistance under the MsfPablo persona.


lock_api/src/mutex.rs
```diff
@@ -872,6 +872,58 @@ impl<R: RawMutex, T: ?Sized> Drop for ArcMutexGuard<R, T> {
...
+///
+/// // Leak the mutex so the mapped guard can hold a `'static` reference.
+/// let m: &'static Mutex<NotSyncRawMutex, (u8, u8)> =
+///     Box::leak(Box::new(Mutex::new((0, 0))));
+/// let guard: MappedMutexGuard<'static, NotSyncRawMutex, u8> =
+///     MutexGuard::map(m.lock(), |p| &mut p.1);
+/// // Sending the guard across threads would run `unlock` on the wrong
+/// // thread, racing on the `Cell`-backed `owner_tid`.
+/// std::thread::spawn(move || drop(guard)).join().unwrap();
+/// ```
```
Amanieu, line 926: This entire block of text is unnecessary, please remove it. Apart from that, the changes look good!

Amanieu:
Closing in favor of #533

PR #532, opened 2026-08-17
add RwLock recursive detect

### LINKED ISSUE #438: [Feature Request] panic when user try get recursive lock.
I had notice that rwlock is not a recursive lock:

> This lock uses a task-fair locking policy which avoids both reader and writer starvation. This means that readers trying to acquire the lock will block even if the lock is unlocked when there are writers waiting to acquire the lock. Because of this, attempts to recursively acquire a read lock within a single thread may result in a deadlock.

That is to say, I cannot acquire the read lock and write lock at the same time in one thread.

On Linux, trying to acquire a recursive lock on a non-recursive lock causes the program to crash. Should packing_lot also have this behavior? Crashing the program early can reduce debugging time (especially for problems like deadlocks)

cathaysia:
fix #438 

Hi, sorry for the delay in submitting this PR.

This PR adds logic under the `detect_deadlock` feature to detect possible deadlocks with read-write locks.

I use a TLS HashMap, with the `self` pointer address as the key and an enum `RwLockMode` for the lock state. During lock, unlock, and other phases, it attempts to check whether a deadlock might occur.

By default, a detected deadlock only prints a warning. I've added a new environment variable, `PARKING_LOT_PANIC_ON_DEADLOCK`, which will cause a panic if set to `true`.

I wrote a test matrix, and here is the actually result runs on macos:

| Lock sequence | Behavior without detection | With deadlock detection | Classification |
|---|---|---|---|
| `read -> read` | Acquired | Panic | ⚠️ Potential deadlock |
| `read -> read` with a waiting writer | Timed out | Panic | ❌ Deadlock |
| `read -> read_recursive` | Acquired | No panic | ✅ Safe |
| `read -> write` | Timed out | Panic | ❌ Deadlock |
| `read -> upgradable` | Acquired | No panic | ✅ Safe |
| `write -> read` | Timed out | Panic | ❌ Deadlock |
| `write -> write` | Timed out | Panic | ❌ Deadlock |
| `upgradable -> read` | Acquired | No panic | ✅ Safe |
| `upgradable -> write` | Timed out | Panic | ❌ Deadlock; use `upgrade` |
| `upgradable -> upgradable` | Timed out | Panic | ❌ Deadlock |
| `upgradable -> upgrade` | Acquired | No panic | ✅ Safe transition |
| `upgrade -> drop -> upgradable` | Acquired | No panic | ✅ Safe |
| `upgrade -> downgrade -> read` | Acquired | Panic | ⚠️ Potential deadlock, equivalent to `read -> read` |
| `write -> downgrade(read) -> upgradable` | Acquired | No panic | ✅ Safe |
| `write -> downgrade_to_upgradable -> read` | Acquired | No panic | ✅ Safe |
| `upgradable -> downgrade(read) -> upgradable` | Acquired | No panic | ✅ Safe |

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,10 @@ parking_lot = { version = "0.12", features = ["nightly"] }
 The experimental deadlock detector can be enabled with the
 `deadlock_detection` Cargo feature.
 
+With this feature enabled, recursive `RwLock` acquisitions are reported as
+warnings on stderr. Set `PARKING_LOT_PANIC_ON_DEADLOCK` to `1`, `true`, `yes`,
+or `on` to panic when such an acquisition is detected.
+
 To allow sending `MutexGuard`s and `RwLock*Guard`s to other threads, enable the
 `send_guard` option.
 
--- a/core/src/parking_lot.rs
+++ b/core/src/parking_lot.rs
@@ -1130,6 +1130,75 @@ pub mod deadlock {
         deadlock_impl::release_resource(_key);
     }
 
+    /// The kind of access held on an `RwLock` for recursive-lock detection.
+    #[derive(Clone, Copy, Debug)]
+    pub enum RwLockMode {
+        /// An ordinary shared read access.
+        Read,
+        /// An explicitly recursive shared read access.
+        RecursiveRead,
+        /// An upgradable shared access.
+        Upgradable,
+        /// An exclusive write access.
+        Write,
+    }
+
+    /// Checks whether acquiring an `RwLock` may deadlock with a lock held by
+    /// the current thread.
+    ///
+    /// With `deadlock_detection` disabled this is a no-op. With it enabled,
+    /// a warning is printed to stderr when a recursive acquisition is found.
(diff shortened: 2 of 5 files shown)


cathaysia:
🤖 This pull request will be closed in 15 days if there are no new non-automation comments.

tracked by [#70](https://github.com/cathaysia/cathaysia/actions/runs/33726737698)

Amanieu:
I'd like to get #533 merged first, since that has several fixes for RwLock deadlock detection. However I'm likely going to be busy for the next few weeks, and there are still several issues that need to be solved there.

cathaysia:
Hi @Amanieu,

Thank you for the update!

I usually prefer not to track long-running PRs over an extended period. Considering that this PR was primarily written with the help of an LLM, I've decided to close it for now to avoid leaving it open indefinitely.

I will keep the branch in my fork intact. If you or anyone else is interested in this feature down the road, hopefully the branch and test matrix might still be of some help or reference.

Thanks again for your time and for maintaining parking_lot!

